PR #14, opened 2019-09-09
Load data from multinet

JackWilb:


JackWilb:
The functions don't load all the data from the tables right now and the graph doesn't display

JackWilb:
I'm going to merge this since it's technically able to pull data from multinet now. We'll need some more development (#17, #18, #19. etc.)

Issue #18, opened 2019-09-23
Fix the node highlighting


JackWilb:
Clicking or searching nodes selects all of them, not just the one we're looking for

PR #21, opened 2019-09-23
Implement tests

### LINKED ISSUE #5: Implement tests
We need unit, integration, and UI tests

JackWilb:
Add unit and integration tests for the app. We should expect that some of these will fail since the app isn't in a completed state (for example selecting nodes). Closes #5

There is a partial fix here for #18. Neighbors will not be fixed


JackWilb:
Going to merge since this has a good set up for tests. We'll need to expand the suite and add in integration/ e2e tests

PR #24, opened 2019-09-25
Refactor code

JackWilb:
Closes #25 

TODOS:

- [ ] Remove commented out code (not fixed, #25 still open)
- [ ] Reduce redundant code by moving it into functions (not fixed, #25 still open)
- [ ] Better comments (not fixed, #25 still open)
- [x] Convert to type script (opened #27 )
- [x] Fix the name spaces (this is ongoing but has mostly been moved to a better namespace format)
- [x] Remove evals (closes #16)
- [x] Remove console.logs
- [x] Stop all the errors (have stopped all apart from the export ones since they're used for testing)
- [x] Fix margin issues
- [x] Fix dragging of nodes to constrain them to the svg area (closes #20)
- [x] Color nodes by table (closes #17 )

PR #26, opened 2019-09-26
Add more tests

### LINKED ISSUE #23: Add more tests
The test suite is sparse.

JackWilb:
Add more tests to cover more code. This should be merged after #24 to make sure that the code is in a better format. This will also add puppeteer and should be able to run a local instance of the server with docker, api, and webserver. This will require using environment variables to specify the multinet call address.

Closes #23


JackWilb:
going to merge this in so the e-2-e tests run each time. This will be useful for our code refactoring

Issue #32, opened 2019-11-04
Implement the control panel buttons and switches into the nodelink


JackWilb:
The control panel currently has no utility currently. It need to allow users to show/hide whatever they want, resize elements, etc.

JackWilb:
Add a switch for link forces

PR #34, opened 2019-11-11
Implement some of the control panel functionality

JackWilb:
Partial solution to #32. This PR will fix the control panel pieces that don't influence the types of node drawn (e.g. nested bars).


JackWilb:
When restoring a config, the links don't get drawn properly

JackWilb:
A few issues remain. Toggling settings in the control panel redraws the nodes a little closer to the center and sometimes restoring breaks the links (they don't track with the nodes anymore). The links issue could be due to the data changing in the back so d => doesn't reference the link any more.

PR #41, opened 2019-12-18
Add a workflow that triggers an action in multinet-deploy

JackWilb:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,18 @@
...
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    name: Build and push to deploy
+    runs-on: ubuntu-latest
+    steps:
+      - run: "curl -XPOST -u \"${{secrets.USERNAME}}:${{secrets.TOKEN}}\" -H \"Accept: application/vnd.github.everest-preview+json\"  -H \"Content-Type: application/json\" https://api.github.com/repos/multinet-app/multinet-deploy/dispatches --data '{\"event_type\":\"build_application\", \"client_payload\": {\"repo\": \"faux-app\"}}'"
```
waxlamp:
```suggestion
      - run: "curl -XPOST -u \"${{secrets.USERNAME}}:${{secrets.TOKEN}}\" -H \"Accept: application/vnd.github.everest-preview+json\"  -H \"Content-Type: application/json\" https://api.github.com/repos/multinet-app/multinet-deploy/dispatches --data '{\"event_type\":\"build_application\", \"client_payload\": {\"repo\": \"view-nodelink\"}}'"
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,18 @@
+
+name: Build
+on:
+  pull_request:
+    branches:
+      - master
```
waxlamp: I think this needs to be changed to something like

```yaml
on:
  push:
    branches:
      - master
```

There's no "merged PR" condition, but I believe this will cover both merged PRs, and direct pushes to master (which we don't normally do but might be useful in an emergency).

waxlamp (changes requested):

.github/workflows/deploy.yml
JackWilb: We must have noticed this at the same time. I updated it in 540b6cb

waxlamp: Great minds :joy:

.github/workflows/deploy.yml
JackWilb: This already existed below so I just removed the PR condition in c4caf45

waxlamp: Ah I didn't even notice. Cool.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,13 @@
+
```
waxlamp:
```suggestion

```

waxlamp (approved):
If you just accept the suggested whitespace change (delete blank line) then this is good to go.

PR #43, opened 2019-12-19
Re-implement the testing framework

### LINKED ISSUE #39: Add testing infrastucture back in with puppeteer
Got lost in conversion to vue

JackWilb:
Closes #39. 

This PR adds a couple tests just as a proof of concept and a template that we can copy across to other apps. The test suite will need to be fleshed out in more PRs


JackWilb:
Here's an example I was working from that used vuetify. https://github.com/adityaekawade/Vue.js-unit-test-with-jest/blob/master/tests/unit/Home.spec.js

Still having some errors with unknown custom elements such as <v-app>

JackWilb:
We don't need to use vuetify if we do a proper mount so I removed it

waxlamp (approved):
Looks great!

PR #44, opened 2020-01-03
Nested bars

### LINKED ISSUE #33: Allow nested bars + glyphs on the nodes
Re-implement the nested bar view and allow it to be controlled from the control panel.

JackWilb:
Closes #33. 

Todo:
- [x] Move node labels
- [x] Add bars
- [x] Add glyphs
- [x] Make everything resize correctly
- [x] Support zero, one, or multiple bars
- [x] Support zero, one, or multiple glyphs (currently at most 2)
- [x] Map the variables from data to the visuals
- [x] Dynamically assign variables
- [ ] Restrict categorical vars to glyphs and numeric to bars
- [ ] Automatically bump up the size of the node if it's too small to render what's requested
- [ ] Fix bug where you can switch to circle with renderNested set to true (only fixed when clicking the radio button, not the text)

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -175,6 +175,7 @@ export default {
             </v-card-subtitle>
 
             <v-select
+              v-if="renderNested"
               v-model="nestedBarVariables"
               :items="variableList"
               label="Bar Variables"
@@ -185,6 +186,7 @@ export default {
             />
 
             <v-select
+              v-if="renderNested"
               v-model="nestedGlyphVariables"
               :items="variableList"
               label="Glyph Variables"
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -116,7 +116,7 @@ export default {
           <v-card-text>
             <v-card-subtitle class="pb-0 pl-0">Marker Type</v-card-subtitle>
             <v-radio-group v-model="nodeMarkerType">
-              <v-radio name="active" label="Circle" value="Circle"></v-radio>
+              <v-radio name="active" label="Circle" value="Circle" @click="renderNested = false; nodeMarkerType = 'Circle'"></v-radio>
               <v-radio name="active" label="Rectangle" value="Rectangle"></v-radio>                
             </v-radio-group>
 
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -130,6 +130,7 @@ export default {
               <v-switch
                 class="ma-0"
                 v-model="renderNested"
+                :disabled="nodeMarkerType === 'Circle'"
                 hide-details
               />
             </v-card-subtitle>
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -191,7 +191,9 @@ export default {
               :items="variableList"
               label="Glyph Variables"
               multiple
+              counter=2
               chips
+              deletable-chips="true"
               hint="Choose the variables you'd like to model as glyphs"
               persistent-hint
             />
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -37,6 +37,8 @@ export default {
       graph: null,
       selectNeighbors: true,
       renderNested: false,
+      nestedBarVariables: [],
+      nestedGlyphVariables: [],
       labelVariable: "_key",
       colorVariable: "table",
     };
@@ -172,6 +174,26 @@ export default {
               />
             </v-card-subtitle>
 
+            <v-select
+              v-model="nestedBarVariables"
+              :items="variableList"
+              label="Bar Variables"
+              multiple
+              chips
+              hint="Choose the variables you'd like to model as bars"
+              persistent-hint
+            />
+
+            <v-select
+              v-model="nestedGlyphVariables"
+              :items="variableList"
+              label="Glyph Variables"
+              multiple
+              chips
+              hint="Choose the variables you'd like to model as glyphs"
+              persistent-hint
+            />
+
             <v-divider class="mt-4" />
 
             <v-card-subtitle class="pb-0 px-0" style="display: flex; align-items: center; justify-content: space-between">
@@ -221,6 +243,8 @@ export default {
               renderNested,
               labelVariable,
               colorVariable,
+              nestedBarVariables,
+              nestedGlyphVariables,
             }"
             @restart-simulation="hello()"
             />
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -36,6 +36,7 @@ export default {
       workspace: null,
       graph: null,
       selectNeighbors: true,
+      renderNested: false,
     };
   },
 
@@ -98,6 +99,8 @@ export default {
               <v-radio name="active" label="Rectangle" value="Rectangle"></v-radio>                
             </v-radio-group>
 
+            <v-divider class="mt-4" />
+
             <v-card-subtitle class="pb-0 pl-0">Marker Size</v-card-subtitle>
             <v-slider
               v-model="nodeMarkerSize"
@@ -119,6 +122,17 @@ export default {
               inverse-label
               hide-details
             />
+
+            <v-divider class="mt-4" />
+
+            <v-card-subtitle class="pb-0 px-0" style="display: flex; align-items: center; justify-content: space-between">
+              Render Nested Elements
+              <v-switch
+                class="ma-0"
+                v-model="renderNested"
+                hide-details
+              />
+            </v-card-subtitle>
 
             <v-divider class="mt-4" />
 
@@ -166,6 +180,7 @@ export default {
               nodeMarkerType: nodeMarkerType,
               nodeFontSize,
               selectNeighbors,
+              renderNested,
             }"
             @restart-simulation="hello()"
             />
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -31,6 +31,7 @@ export default {
         links: []
       },
       nodeMarkerSize: 50,
+      nodeMarkerType: "Circle",
       nodeFontSize: 14,
       workspace: null,
       graph: null,
@@ -91,6 +92,12 @@ export default {
           <v-card-title class="pb-6">MultiNet Node Link Controls</v-card-title>
 
           <v-card-text>
+            <v-card-subtitle class="pb-0 pl-0">Marker Type</v-card-subtitle>
+            <v-radio-group v-model="nodeMarkerType">
+              <v-radio name="active" label="Circle" value="Circle"></v-radio>
+              <v-radio name="active" label="Rectangle" value="Rectangle"></v-radio>                
+            </v-radio-group>
+
             <v-card-subtitle class="pb-0 pl-0">Marker Size</v-card-subtitle>
             <v-slider
               v-model="nodeMarkerSize"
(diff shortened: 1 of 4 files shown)


JackWilb:
We're going to need to guess which variables are numeric vs categorical but let the user select which is which for when our assumption is wrong.

JackWilb:
@waxlamp This should be pretty much ready to go. With your approval, I'll make the remaining TODOs into issues and merge. I think enough of the core functionality is here that we can go, but let me know.

waxlamp:
> We're going to need to guess which variables are numeric vs categorical but let the user select which is which for when our assumption is wrong.

There are libraries that can help with this. For instance, https://github.com/vega/datalib was developed during the XDATA project and might be of use here (though it doesn't seem to be in active use; I'm not sure if something else deprecated it or what).

waxlamp:
Here's a list of things I noticed while playing with this:
- node color option doesn't work in rectangle mode
- upper limit of 4 bars seems too low
- what colors are being used for the bars?
  - more generally, it feels like the bars should be set a bit closer together, and all be the same color
  - or, a linear colormap could be used to redundantly encode the value (i.e., in addition to the bar height)
- setting variables to use for the bars, then setting a glyph variable causes the bars to disappear
- there's no limit to how many glyph vars can be setting, resulting in an overflow
- circle mode no longer seems necessary
  - the color variable can be used in rectangle mode
  - by dumping circle mode, we could also dump the UI logic requiring rectangle mode before turning on the nested elements, and the trouble resulting from reverting to circle mode after activating the nested elements
  - the mission of this visualization is to handle multivariate networks, so the base circle case is less valuable
- we may want some kind of detail view that pops up more traditional plots for the data on a selected node, but this is something to discuss further in a separate issue/PR

I'll do a code review separately, just wanted to get some initial reactions on paper.

This is looking really good!

JackWilb:
> Here's a list of things I noticed while playing with this:
> 
>     * node color option doesn't work in rectangle mode

It does! You just can't control node color when nested render is turned on. I think this is a suitable default.

>     * upper limit of 4 bars seems too low

The only upper limit is what can be rendered. You can get to 9 bars with the max marker size of 100.

>     * what colors are being used for the bars?

It's assigned by variable name using one of the color scales.

>       * more generally, it feels like the bars should be set a bit closer together, and all be the same color

Spacing is something we can change, but since it's design decision I think we should open a new PR for that.

>       * or, a linear colormap could be used to redundantly encode the value (i.e., in addition to the bar height)

Sure. This is something Alex could weigh in on.

> 
>     * setting variables to use for the bars, then setting a glyph variable causes the bars to disappear

This happens because the width of the bars becomes zero or close to it when you have many bars. I agree this isn't ideal behavior but reducing the number of bars remedies this. 

>     * there's no limit to how many glyph vars can be setting, resulting in an overflow

We should set a sensible limit. Maybe 3 or 4?

>     * circle mode no longer seems necessary
>       * the color variable can be used in rectangle mode
>       * by dumping circle mode, we could also dump the UI logic requiring rectangle mode before turning on the nested elements, and the trouble resulting from reverting to circle mode after activating the nested elements
>       * the mission of this visualization is to handle multivariate networks, so the base circle case is less valuable

Agreed. We should ask Alex if he agrees here.

>     * we may want some kind of detail view that pops up more traditional plots for the data on a selected node, but this is something to discuss further in a separate issue/PR
> 
> 
> I'll do a code review separately, just wanted to get some initial reactions on paper.
> 
> This is looking really good!

waxlamp:
> > Here's a list of things I noticed while playing with this:
> > ```
> > * node color option doesn't work in rectangle mode
> > ```
> 
> It does! You just can't control node color when nested render is turned on. I think this is a suitable default.
> 

For me, the color of the node completely disappears when I switch on the nested bars. In my view, the color should remain visible behind the node label (and in this situation, there wouldn't be anything stopping the user from changing the color setting in this mode).

> > ```
> > * upper limit of 4 bars seems too low
> > ```
> 
> The only upper limit is what can be rendered. You can get to 9 bars with the max marker size of 100.
> 

I see. If I didn't understand how that works, then neither will the average user. We'll have to come up with some sort of affordance to make sure users don't simply think the visualization broke.

> > ```
> > * what colors are being used for the bars?
> > ```
> 
> It's assigned by variable name using one of the color scales.
> 

This is probably not the right thing, since mapping the variable name to a bar color doesn't carry a lot of information that isn't better displayed through a level-of-detail type of presentation. For now I would default all the bars to a single color (a pleasant shade of blue or something like that).

> > ```
> >   * more generally, it feels like the bars should be set a bit closer together, and all be the same color
> > ```
> 
> Spacing is something we can change, but since it's design decision I think we should open a new PR for that.
> 

Ok.

> > ```
> >   * or, a linear colormap could be used to redundantly encode the value (i.e., in addition to the bar height)
> > ```
> 
> Sure. This is something Alex could weigh in on.
> 

Agreed that we should collect opinions on this, but even this idea seems marginally less useful than using a uniform color for the bars.

> > ```
> > * setting variables to use for the bars, then setting a glyph variable causes the bars to disappear
> > ```
> 
> This happens because the width of the bars becomes zero or close to it when you have many bars. I agree this isn't ideal behavior but reducing the number of bars remedies this.
> 

Sure, but that's not a permanent remedy. We have to be able to show a larger number of variables in this view, since that's its core purpose. Let's discuss better solutions in followup issues/PRs.

> > ```
> > * there's no limit to how many glyph vars can be setting, resulting in an overflow
> > ```
> 
> We should set a sensible limit. Maybe 3 or 4?
> 

Another option is to plot these in a growing grid, so they don't end up overflowing in the y direction. But for now setting a limit is probably ok. The same concern I have above about displaying a large number of variables applies here too.

> > ```
> > * circle mode no longer seems necessary
> >   * the color variable can be used in rectangle mode
> >   * by dumping circle mode, we could also dump the UI logic requiring rectangle mode before turning on the nested elements, and the trouble resulting from reverting to circle mode after activating the nested elements
> >   * the mission of this visualization is to handle multivariate networks, so the base circle case is less valuable
> > ```
> 
> Agreed. We should ask Alex if he agrees here.
> 

Good idea. Friday is probably a good time to demo this to the big group, raise any concerns/questions we have, and collect feedback.

Thanks Jack!

src/components/Controls.vue
```diff
@@ -31,10 +31,14 @@ export default {
         links: []
       },
       nodeMarkerSize: 50,
+      nodeMarkerType: "Circle",
```
waxlamp, line 34: Just an idea for the future: rather than a string, you might want to consider using [Symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) to implement a small enum system for the different `nodeMarkerType` values.

src/components/Controls.vue
```diff
@@ -220,10 +269,14 @@ export default {
               app,
               nodeMarkerHeight: nodeMarkerSize,
               nodeMarkerLength: nodeMarkerSize,
+              nodeMarkerType: nodeMarkerType,
```
waxlamp:
```suggestion
              nodeMarkerType,
```

src/components/NodeLink/NodeLink.vue
```diff
@@ -69,6 +73,18 @@ export default {
         edgeWidthKey: undefined,
       })
     },
+    renderNested: {
+      type: Boolean,
+      default: false
+    },
+    nestedBarVariables: {
+      type: Array[String],
```
waxlamp: I don't think this is valid JavaScript. To accomplish the intent you would change this to `type: Array,`, and then add a `validator` section that looks something like

```javascript
validator: (prop) => prop.every((item) => typeof item === 'string'),
```

The other option here is to use TypeScript, which makes it easier to do this sort of check without the runtime validator.

src/components/NodeLink/NodeLink.vue
```diff
@@ -69,6 +73,18 @@ export default {
...
+      default: () => []
```
waxlamp: Why not just `[]` here?

src/components/NodeLink/NodeLink.vue
```diff
@@ -69,6 +73,18 @@ export default {
...
+    },
+    nestedGlyphVariables: {
+      type: Array[String],
+      default: () => []
```
waxlamp: (same question as above)

src/components/NodeLink/NodeLink.vue
```diff
@@ -69,6 +73,18 @@ export default {
...
+    renderNested: {
+      type: Boolean,
+      default: false
+    },
+    nestedBarVariables: {
+      type: Array[String],
+      default: () => []
+    },
+    nestedGlyphVariables: {
+      type: Array[String],
```
waxlamp: (same comment as above)

waxlamp (approved):
This is looking pretty solid architecture-wise. If the idea is to iterate on the vis content, and discuss the options further (such as my opinion that we should get rid of circle mode) then I think this is good to merge now.

I included a couple of suggestions in the comments but I don't think any of them are absolutely critical. I did also notice several prop type errors in the console, and when the bar glyphs disappear, it tends to be because the code computes negative values for their geometry--these are things we should fix in future iterations.

src/components/Controls.vue
JackWilb: Agreed. I'll keep this in mind for future updates

src/components/NodeLink/NodeLink.vue
JackWilb: I was trying to be more specific with types. Are you in favor of just switching all these to Array instead of Array[String] since it will reduce complexity?

src/components/NodeLink/NodeLink.vue
JackWilb: `Type of the default value for 'nestedBarVariables' prop must be a function` if it's not a function

src/components/NodeLink/NodeLink.vue
JackWilb: Same the bar version

src/components/NodeLink/NodeLink.vue
waxlamp: It's good to be specific with types, but what I'm saying is, `Array[String]` is not a valid JavaScript type. Setting it to `Array` *is* valid, though not very specific; adding that `validator` to the prop definition will let you validate that every element of the prop is in fact a string.

This may explain why Vue is telling you that your default value needs to be a function (though that's just a guess). Technically, `Array[String]` evaluates to `undefined`, so maybe that's somehow making Vue infer that the prop is a function type. (I wouldn't put money on this explanation.)

The better approach would be to set the type as `Array as string[]` but that would require TypeScript which of course is a major upgrade to make. So at least for now I'd try out the `Array`+`validator` approach.

JackWilb: Should be fixed in the recent commits

src/components/NodeLink/NodeLink.vue
JackWilb: Still getting this with the new validator and type

src/components/NodeLink/NodeLink.vue
JackWilb: Resolving since it's the same as above

src/components/NodeLink/NodeLink.vue
JackWilb: Resolving since it's the same as above

src/components/NodeLink/NodeLink.vue
waxlamp: Huh I didn't realize this warning was coming from the linter. I guess this is to force a fresh copy of a mutable object (like a list) to be constructed to avoid weird problems with reference sharing. So follow the linter and keep the `default` value as is (but still use the type and validator we discussed before).

PR #50, opened 2020-01-09
Re-implement dynamic labels and add node coloring

### LINKED ISSUE #46: Re-implement label switching

### LINKED ISSUE #49: Implement node coloring by attributes
These would include those passed through from the multinet api including the table name and any other variable in the data

JackWilb:
Closes #46
Closes #49

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -37,9 +37,24 @@ export default {
       selectNeighbors: true,
       labelVariable: "_key",
       colorVariable: "table",
-      variableList: [],
-      colorVariableList: [],
     };
+  },
+
+  computed: {
+    variableList() {
+      if (typeof this.graphStructure.nodes[0] !== 'undefined') {
+        return Object.keys(this.graphStructure.nodes[0]) 
+      } else {
+        return []
+      }
+    },
+    colorVariableList() {
+      if (typeof this.graphStructure.nodes[0] !== 'undefined') {
+        return Object.keys(this.graphStructure.nodes[0]).concat(["table"]) 
+      } else {
+        return []
+      }
+    }
   },
 
   /**
@@ -59,8 +74,9 @@ export default {
     this.provenance = provenance;
     this.workspace = workspace;
     this.graph = graph;
-    this.variableList = Object.keys(this.graphStructure.nodes[0]);
-    this.colorVariableList = Object.keys(this.graphStructure.nodes[0]).concat(["table"]);
+    console.log(this.properties)
+    // this.variableList = Object.keys(this.graphStructure.nodes[0]);
+    // this.colorVariableList = Object.keys(this.graphStructure.nodes[0]).concat(["table"]);
   },
 
   methods: {
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -74,9 +74,6 @@ export default {
     this.provenance = provenance;
     this.workspace = workspace;
     this.graph = graph;
-    console.log(this.properties)
-    // this.variableList = Object.keys(this.graphStructure.nodes[0]);
-    // this.colorVariableList = Object.keys(this.graphStructure.nodes[0]).concat(["table"]);
   },
 
   methods: {
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -35,8 +35,10 @@ export default {
       workspace: null,
       graph: null,
       selectNeighbors: true,
-      labelVariable: null,
+      labelVariable: "_key",
+      colorVariable: "_key",
       variableList: [],
+      colorVariableList: [],
     };
   },
 
@@ -58,6 +60,7 @@ export default {
     this.workspace = workspace;
     this.graph = graph;
     this.variableList = Object.keys(this.graphStructure.nodes[0]);
+    this.colorVariableList = Object.keys(this.graphStructure.nodes[0]).concat(["table"]);
   },
 
   methods: {
@@ -127,6 +130,15 @@ export default {
 
             <v-divider class="mt-4" />
 
+            <v-select 
+              v-model="colorVariable"
+              label="Color Variable"
+              :items="colorVariableList"
+              :options="colorVariableList"
+            />
+
+            <v-divider class="mt-4" />
+
             <v-card-subtitle class="pb-0 px-0" style="display: flex; align-items: center; justify-content: space-between">
               Autoselect neighbors
               <v-switch
@@ -170,6 +182,8 @@ export default {
               nodeMarkerLength: nodeMarkerSize,
               nodeFontSize,
               selectNeighbors,
+              labelVariable,
+              colorVariable,
             }"
             @restart-simulation="hello()"
             />
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -36,7 +36,7 @@ export default {
       graph: null,
       selectNeighbors: true,
       labelVariable: "_key",
-      colorVariable: "_key",
+      colorVariable: "table",
       variableList: [],
       colorVariableList: [],
     };
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -35,6 +35,8 @@ export default {
       workspace: null,
       graph: null,
       selectNeighbors: true,
+      labelVariable: null,
+      variableList: [],
     };
   },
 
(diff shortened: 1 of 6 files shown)


JackWilb:
@waxlamp I'm having some issues with this property calling out to updateVis on updated. Can you take a look and show me what I'm doing wrong?

JackWilb:
@waxlamp This is ready for review. I also added node coloring to this PR using a similar mechanism

src/components/Controls.vue
```diff
@@ -35,6 +35,10 @@ export default {
       workspace: null,
       graph: null,
       selectNeighbors: true,
+      labelVariable: "_key",
+      colorVariable: "_key",
```
waxlamp: Maybe we can leave these with a default value of `null`, and add an option to each dropdown called `(none)`, and use that to avoid placing a label or a special color at all. This might be more important for the color field, because while _key is possibly appropriate for a default label, it definitely is the wrong choice for a color.

src/components/Controls.vue
```diff
@@ -55,6 +59,8 @@ export default {
     this.provenance = provenance;
     this.workspace = workspace;
     this.graph = graph;
+    this.variableList = Object.keys(this.graphStructure.nodes[0]);
+    this.colorVariableList = Object.keys(this.graphStructure.nodes[0]).concat(["table"]);
```
waxlamp: These should probably be computed variables.

src/components/NodeLink/functionUpdateVis.js
```diff
@@ -175,11 +175,13 @@ function makeSimulation() {
...
-  const { attributes, colorClasses, nodeColors } = this;
-  if (attributes.nodeFill === "table") {
-    const index = colorClasses.findIndex(d => { return d === node.id.split("/")[0] }) % 5
+function nodeFill(node, nodeClasses) {
+  const { colorClasses, nodeColors, colorVariable } = this;
+  if (colorVariable === "table") {
+    const index = colorClasses.findIndex(d => { return d === node.id.split("/")[0] }) % nodeColors.length
     return nodeColors[index]
+  } else {
+    return nodeColors[nodeClasses.findIndex(d => { return d === node[colorVariable]}) % nodeColors.length]
```
waxlamp: Instead of these linear searches, why not just use d3 colormapping utilities?

waxlamp (approved):
I'm approving this because it seems to work great, and the code changes are pretty straightforward (nice work), but I would like you to address my comments and/or file issues as appropriate.

src/components/Controls.vue
JackWilb: I think we should leave the label as _key since it's strange if no label renders.

For colorVariable, we could default to table which is most likely going to be the node type for any given data. Changing to null also seems to be an appropriate solution

src/components/Controls.vue
JackWilb: Doesn't it only make sense to have these be computed if they update? We don't expect that the graphStructure will change though, right?

src/components/Controls.vue
waxlamp: Oh, yeah, table is a good choice for the color option.

src/components/Controls.vue
waxlamp: Computed is good for any value that is a function of other props or data props. In this case, `graphStructure` is a prop, so using a computed prop for `variableList` and `colorVariableList` is the right thing to do.

You're right that we don't expect `graphStructure` to change now, but if in the future we redesign things so that it *does* change (for instance, if we were to add a dropdown allowing us to switch which graph is being visualized, similar to how evo-bio does it), then this code will become incorrect.

One good thing about computed properties is that Vue "caches" them for you transparently. So there shouldn't be any performance impact to using them here.

waxlamp: In fact, due to the argument I just made, I'd say we should definitely transfer these to computed properties before merging.

src/components/Controls.vue
JackWilb: Fixed in a6f9208

src/components/Controls.vue
JackWilb: Fixed in 326673f

src/components/NodeLink/functionUpdateVis.js
JackWilb: Should be fixed in 63e88a2

src/components/NodeLink/functionUpdateVis.js
```diff
@@ -242,8 +236,15 @@ function updateVis() {
     .attr("rx", () => nodeMarkerLength / 2)
     .attr("ry", () => nodeMarkerHeight / 2);
 
+  let nodeClasses = colorVariable === "table" ? [...new Set(graphStructure.nodes.map(value => value[colorVariable]))] : 
+    [...new Set(graphStructure.nodes.map(value => value["id"].split("/")[0]))];
+
+  let nodeColorScale = d3.scaleOrdinal()
+    .domain(nodeClasses)
+    .range(nodeColors);
```
waxlamp: This is fine--it gives you very fine control over the colors and how they are mapped--but it would be simpler to do something like

```javascript
let nodeColorScale = d3.scaleOrdinal(d3.schemeCategory10)
```

When you create a D3 color scale like this, you can just throw values at it and get colors back, and they will be consistent based on what order individual values are invoked. Then you wouldn't have to recompute the list of classes as you do here. And since this color scale could be used for the lifetime of the component, you could even just make it a member of the components data props.

This is mainly just for information. The code is fine as you have it, but could be simplified (while yielding that fine grained control over the colors) to something like what I suggested.

waxlamp (approved):

src/components/NodeLink/functionUpdateVis.js
JackWilb: So you don't have to specify a specific domain? It just maps inputs (consistently) to the same values?

waxlamp: Yeah, it'll map each new input it encounters to the next color (or something along those lines), and then each old input will remain mapped to whatever color came back the first time.

JackWilb: Okay. I'll move it over to the vue component props so it doesn't recompute each time

JackWilb: Fixed in a286cff

src/components/NodeLink/NodeLink.vue
```diff
@@ -23,9 +23,13 @@ export default {
...
     },
-    nodeLabel: {
+    labelVariable: {
       type: String,
-      default: "id"
+      default: "_key"
+    },
+    colorVariable: {
+      type: String,
+      default: "_key"
```
waxlamp: This should be

```suggestion
      default: "table"
```

right?

src/components/NodeLink/NodeLink.vue
```diff
@@ -92,30 +95,33 @@ export default {
       straightEdges: false,
       // distinguish a drag from a click
       wasDragged: false,
+      nodeColorScale: d3.scaleOrdinal(d3.schemeCategory10),
```
waxlamp, line 98: :tada:

src/components/NodeLink/functionUpdateVis.js
```diff
@@ -243,7 +237,7 @@ function updateVis() {
     .attr("ry", () => nodeMarkerHeight / 2);
 
   node.select('.node')
-    .style("fill", d => this.nodeFill(d))
+    .style("fill", d => nodeColorScale(d[colorVariable]))
```
waxlamp: :tada: :tada:

waxlamp (approved):
Looks great--feel free to merge after accepting the suggested change.

src/components/NodeLink/functionLoadVis.js
```diff
@@ -54,15 +54,16 @@ export function loadVis() {
...
-    const table = node.id.split("/")[0];
+  // for (const node of this.graphStructure.nodes) {
+  //   // Get the table name
+  //   const table = node.id.split("/")[0];
+
+  //   // If we haven't seen it, push it to the colorClasses array
+  //   if (!this.colorClasses.includes(table)) {
+  //     this.colorClasses.push(table);
+  //   }
+  // }
```
waxlamp: Remove this code.

waxlamp (approved):

PR #51, opened 2020-01-10
Add link attributes by allowing link width and color to be controlled by the user

### LINKED ISSUE #47: Implement link attributes
For now this will likely be restricted to stroke and color

JackWilb:
Closes #47

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -59,7 +59,7 @@ export default {
     },
     linkVariableList() {
       if (typeof this.graphStructure.links[0] !== 'undefined') {
-        return Object.keys(this.graphStructure.links[0]).concat(["table"]) 
+        return Object.keys(this.graphStructure.links[0]).concat(["table", null]) 
       } else {
         return []
       }
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -37,6 +37,7 @@ export default {
       selectNeighbors: true,
       labelVariable: "_key",
       colorVariable: "table",
+      linkWidthVariable: null,
     };
   },
 
@@ -51,6 +52,13 @@ export default {
     colorVariableList() {
       if (typeof this.graphStructure.nodes[0] !== 'undefined') {
         return Object.keys(this.graphStructure.nodes[0]).concat(["table"]) 
+      } else {
+        return []
+      }
+    },
+    linkVariableList() {
+      if (typeof this.graphStructure.links[0] !== 'undefined') {
+        return Object.keys(this.graphStructure.links[0]).concat(["table"]) 
       } else {
         return []
       }
@@ -136,7 +144,7 @@ export default {
 
             <v-select 
               v-model="linkWidthVariable"
-              label="Label Variable"
+              label="Link Width Variable"
               :items="linkVariableList"
               :options="linkVariableList"
               />
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -59,7 +59,7 @@ export default {
     },
     linkVariableList() {
       if (typeof this.graphStructure.links[0] !== 'undefined') {
-        return Object.keys(this.graphStructure.links[0]).concat(["table", null]) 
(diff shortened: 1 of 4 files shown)


src/components/NodeLink/functionUpdateVis.js
```diff
@@ -298,11 +300,22 @@ function updateVis() {
...
     .select("path")
-    .style("stroke-width", 3)
-    .style("stroke", edgeColor)
+    .style("stroke-width", d => {
+      if (linkWidthVariable) {
+        return d[linkWidthVariable]
+      } else{
+        return 3
+      }
+    })
```
waxlamp: Rather than using an existing variable directly as the link width, it's more likely that we'll want to map a numeric variable to a width using a [d3 continuous scale](https://github.com/d3/d3-scale#continuous-scales). But this is a good start for the infrastructure to handle that.

src/components/NodeLink/functionUpdateVis.js
```diff
@@ -298,11 +300,22 @@ function updateVis() {
...
+    .style("stroke", d => {
+      if (linkColorVariable !== null) {
+        return linkColorScale(d[linkColorVariable])
+      } else{
+        return "#888888"
+      }
+    })
```
waxlamp, line 319: This is great, though we will also want to support continuous colormaps as well. That can certainly be added in a future PR.

src/components/Controls.vue
```diff
@@ -54,6 +56,13 @@ export default {
       } else {
         return []
       }
+    },
+    linkVariableList() {
+      if (typeof this.graphStructure.links[0] !== 'undefined') {
+        return Object.keys(this.graphStructure.links[0]).concat(["table", null]) 
+      } else {
+        return []
+      }
```
waxlamp: It occurs to me that we may also want to allow the user to set width/color of the links based on data residing on the source or target node. That's something to explore in a future PR. But by the same token, I'm not sure it makes sense to offer `table` as an option here, since currently we only allow a single edge table per graph, nor `source` and `target` by themselves, since those turn out to just be IDs for the nodes connected at either end.

waxlamp (dismissed):
This is a great start. I'm good with merging this if we also open issues related to the stuff I brought up in the comments. Let me know what you think.

src/components/NodeLink/functionUpdateVis.js
JackWilb: Agreed, that's definitely going to be an issue. I just didn't even think of it since I was only using the group var. I think that we should fix this before merging

src/components/NodeLink/functionUpdateVis.js
JackWilb: Opened #57

src/components/Controls.vue
JackWilb: Opened #58 for the first part of your comment. I'll remove table as an option and source and target

JackWilb: Fixed in 2a77ea5

src/components/NodeLink/functionUpdateVis.js
JackWilb: Fixed in aff4fb6

src/components/NodeLink/NodeLink.vue
```diff
@@ -120,10 +129,22 @@ export default {
...
         colorVariable,
       };
-    }
+    },
+    linkWidthScale() {
+      return d3.scaleLinear()
+        .domain(
+          [
+            d3.min(this.graphStructure.links.map(d => d[this.linkWidthVariable])),
+            d3.max(this.graphStructure.links.map(d => d[this.linkWidthVariable]))
```
waxlamp, line 143: It would be good to lift the two dependencies out into a destructuring declaration at the top of the function, just to make it visually more apparent what these dependencies are, but that's a minor style improvement that's not blocking this PR.

But hopefully you've gotten a taste for why computed props are nice to work with :cupid:

waxlamp (approved):
image

Issue #52, opened 2020-01-10
Allow users to define domains for variables


JackWilb:
Allow clamping of numeric variables and category selection for categorical

JackWilb:
This is now working with the brushing and the domain controlling for the scales

PR #53, opened 2020-01-10
Add legend svg that reacts to variables that are being shown

### LINKED ISSUE #35: Re-implement the legend
There is remnant code from the legend that we should activate once the nodes are able to be rendered as nested bars with glyphs

JackWilb:
Closes #35 
Starts toward #52 (with brushing)

TODO:
- [x] Fix the bar and glyph mapping for the on node encoding (#71)
  The bars are not using a proper d3.scale and the glyph encoding will map things from outside the domain to values
- [x] Fix the link color mapping (#72)
  The link color mapping maps things from outside the domain (making it look like it isn't working)
- [x] Update the UI to add node coloring to the drag target SVG (#73)
- [x] Update the UI to add node label to the drag target SVG (#73)
- [x] Make the drag targets more visible (#74)
- [x] Make the nested rendering on/off logic more user friendly (don't require a switch) (#75)
- [x] Abstract out some of the elements so that this component can be reused (#77)
- [x] Make the dragged elements always fit inside the drag target (with a wrap or something)(#76)
- [x] Resize the drag targets based on the amounts of items in them (#76)
- [x] Limit the d3 imports to just the pieces that we're using (#70)


JackWilb:
@waxlamp This is a very rough first pass that we probably shouldn't merge until next week, starting on the 17th. I just wanted to get your feedback and advice for how to improve this since there are a lot of interactive elements that have to be shown

waxlamp:
Looks like a good start. The biggest issue right now is that the visual itself doesn't quite work yet:
legend
My recommendation here would be to provide more space for everyone:
- Lay things out vertically instead of horizontally, giving the labels room to be visible
- If that causes too much scrolling, consider splitting the different sections into vuetify tabs

Codewise, the legend component itself seems to be too long (e.g., GitHub won't show the diff for the added file by default) but I'm not sure there's much that can be done about it. I haven't yet found a very good solution for mixing Vue components with a D3 visualization core. I think without some revolutionary way of thinking about such components, they will just be verbose like this.

That being said, I do have some ideas we could try. (I worry, in general, about the technical debt we're accruing by *not* thinking about this sort of thing early on.) One thing is, since this component has a pretty static structure, we can probably dispense with the `d3.append()` calls in the setup code in favor of just laying out the SVG's top-level structure directly in the template. We could also dump a lot of the D3 logic for managing the individual items by using `v-for` directives in the template as well. This is especially apt here because there's no animation being used to render the legend. A lot of code would probably just go away if we did this: no more need for watchers, no more need for that ugly "properties" hack that is necessary in Vue/D3 components in general, and the template could handle all of the D3-join logic just like it does in non-D3 Vue components.

(Tagging @subdavis because we've had discussions of this sort before.)

waxlamp:
This is looking pretty good, though the legend items still crash into each other if there are too many of them. What are the actual changes since my last comment?

JackWilb:
No updates recently, I have this on my list for tomorrow

waxlamp:
> No updates recently, I have this on my list for tomorrow

Gotcha, sorry, I thought you had asked me to look at this.

src/components/Controls.vue
```diff
@@ -1,28 +1,20 @@
...
 
-/**
- * Demo Controls.  This component is meant to demo
- * the capabilities of the NodeLink vis.
- *
- * You'll probably want the controls to look or
- * work differently in your app, so it's recommended that you
- * implement the controls yourself.
- */
+import * as d3 from "d3";
```
waxlamp, line 9: You will definitely want to import things from d3 by name going forward.

src/components/NodeLink/Legend.vue
```diff
@@ -0,0 +1,428 @@
...
+        linkWidthScale,
+        multiVariableList,
+        linkVariableList,
+        barVariables,
+        glyphVariables,
+        widthVariables,
+        colorVariables,
+      };
+    },
+  },
```
waxlamp, line 102: Depending on why exactly we're using this mechanism, we may want to just get rid of it. This is error prone, and its use has to be weighed against the benefit of using it.

waxlamp (approved):

src/components/Controls.vue
JackWilb: Opened #70

src/components/NodeLink/Legend.vue
JackWilb: What exactly are the potential bugs that this would cause?

waxlamp: If you forget to add something here that you then depend on downstream, or more immediately, if you put something in the top part but not the bottom part.

In general, it's better to just directly express the dependencies in each function rather than go through a mechanism like this. Originally I had something like this to get around some limitation of Vue that I probably shouldn't have been messing with (something like trying to make a side effect reactive by making it depend on a bundle like this, and attaching it to something invisible in the DOM to force updates--it was ugly and wrong).

JackWilb: Makes sense. I'll start to break this up as I work through the other issues then so the dependencies for each function are more apparent.

PR #54, opened 2020-01-10
Fix the graph export

### LINKED ISSUE #45: The export graph button is broken
Currently it gives the right prompt but the data just says undefined. Clearly there is just an issue with the data object that gets outputted

JackWilb:
Closes #45.

I was also going to add an import graph button but that doesn't really make sense with our ecosystem and the validation that would require

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -91,11 +91,16 @@ export default {
 
     exportGraph() {
       const a = document.createElement('a');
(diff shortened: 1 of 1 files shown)


waxlamp:
Sorry I didn't get to this earlier. Code looks good. And as for importing graphs, there may be a use case like that that comes up, in particular if we are able to use tools like Origraph to do some sort of graph editing for the data we want to have in Multinet. So it's good to at least have the concept on your mind.

PR #64, opened 2020-02-05
Allow for `host` parameter to connect to different Multinet instances

waxlamp:


JackWilb (commented):
What should the value of host be to use localhost?

JackWilb (approved):
This works:
host=http://localhost:5000/api

waxlamp:
> What should the value of host be to use localhost?

Depends on where you're running your server :smile_cat:

PR #79, opened 2020-04-02
Update the glyph color scale and don't map from outside the domain

### LINKED ISSUE #71: Use d3 scales for the bars
Currently they're manually computed but to use the legend properly (setting domains, etc.) we should be using the d3.scales

### LINKED ISSUE #72: Stop d3 from mapping things outside the domain to nested glyph colors
Right now, even items that are not in the domain are mapped to "the next available item in the range". This is not intended behavior for our app. There may be a way to stop this with d3 or we might have to define a function to do it

JackWilb:
Will close #72 
Will close #71 
Only maps glyph color scale to the glyphs if the category is in the legend brush.
Move the node bar size calculations to a d3.scale

TODO:
- [ ] Make the glyph colors idempotent even as the brush moves


JackWilb:
Closing, because it will be hard to integrate these changes after the re-architect

PR #80, opened 2020-04-16
Make nested rendering just one toggle instead of 2 actions

### LINKED ISSUE #75: Refactor the nested rendering logic to make selecting it more simple
Roni had suggested maybe using a selector that goes between topology only and attributes. The labels might change but this seems like a good idea

JackWilb:
Closes #75

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -124,7 +124,7 @@ export default {
 
           <v-card-text>
             <v-card-subtitle class="pb-0 pl-0" style="display: flex; align-items: center; justify-content: space-between">
-              Render Nested
+              Display charts
               <v-switch
                   class="ma-0"
                   v-model="renderNested"
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -24,7 +24,6 @@ export default {
         links: []
       },
       nodeMarkerSize: 50,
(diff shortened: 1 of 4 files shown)


src/components/Controls.vue
```diff
@@ -124,11 +123,14 @@ export default {
           <v-card-title class="pb-6">MultiNet Node Link Controls</v-card-title>
 
           <v-card-text>
-            <v-card-subtitle class="pb-0 pl-0">Marker Type</v-card-subtitle>
-            <v-radio-group v-model="nodeMarkerType">
-              <v-radio name="active" label="Circle" value="Circle" @click="renderNested = false; nodeMarkerType = 'Circle'"></v-radio>
-              <v-radio name="active" label="Rectangle" value="Rectangle"></v-radio>                
-            </v-radio-group>
+            <v-card-subtitle class="pb-0 pl-0" style="display: flex; align-items: center; justify-content: space-between">
+              Render Nested
```
waxlamp: Let's not use the phrase "render nested", as it's not super informative. Something like:

```suggestion
              Display charts
```

might work better.

waxlamp (changes requested):

waxlamp (approved):

PR #84, opened 2020-06-18
Convert nodelink to typescript

### LINKED ISSUE #27: Convert to typescript
Properly type the functions for faster development and better coding practices

JackWilb:
Closes #27

TODO:
- [x] Open an issue to add more type safety, this is still most plain js at the moment
- [ ] Tests need to be fixed now or with another issue, won't pass CI as it stands (looks like a path error for the types.ts file)


.github/workflows/main.yml
```diff
@@ -30,4 +30,4 @@ jobs:
       run: yarn install
 
     - run: yarn lint
-    - run: yarn test
+    # - run: yarn test
```
waxlamp: The problem that goes away from this change seems fundamental. I think we should try to fix it before merging this.

waxlamp (changes requested):

.github/workflows/main.yml
waxlamp: I think what's happening is that the module system is looking for a `.js` file when it attempts to import `@/types`. If I change that to `@/types.ts`, then I get a different error, related to Jest not knowing how to run Babel on a typescript file.

I think you need to add TS support to the Vue project itself. I am not quite sure how to do that. Look into it a bit, and I'll consult some experts as well.

waxlamp: Woohoo! With @subdavis's help, I was able to get the tests running (and passing)!

waxlamp (approved):

JackWilb:
Thanks for the help @waxlamp and @subdavis!

PR #87, opened 2020-07-13
Add back the provenance lib

### LINKED ISSUE #55: Re-implement provenance
Use Kiran's package to track state. This should allow undo and redo

### LINKED ISSUE #90: Node-Link: preserve node selection when changing display options
When displaying a graph, clicking nodes highlights the one-hop network around the node.  Nice!  I'd like to request preserving selections on nodes  when an option in the rendering menu is changed.  Currently, changing font size, picking a different attribute for label value, etc. resets all nodes as selected and forces the user to repeat selection operations to get the displayed subnetwork back again.

### LINKED ISSUE #91: Node-Link View Permissions Access
Currently do not have access to node-link view for private workspaces

JackWilb:
Closes #55.
Closes #91.
Closes #90.

TODO:
- [x] Implement visual
- [x] Fix the slowness in dragging nodes
- [x] Check to see that this branch addresses #90 (add to top of desc. if it does)
- [x] Move to trrack, the new name for the provenance lib
- [x] Fix the tests
- [x] Open issue to track adding more/different data into the state object (should we track networks? Are there better ways to track position, etc.)
- [x] Open issue for implementing the provenance visual (so we can move around the provenance graph)


src/components/Controls.vue
```diff
@@ -82,11 +81,11 @@ export default {
...
     }
     this.graphStructure = await loadData(workspace, graph, host);
-    const { provenance, app } = setUpProvenance(this.graphStructure.nodes);
-    this.app = app;
-    this.provenance = provenance;
+    this.provenance = setUpProvenance(this.graphStructure);
     this.workspace = workspace;
     this.graph = graph;
+
+    document.addEventListener('keydown', this.keyDownHandler);
```
waxlamp, line 88: Can you look to see if there's some "Vueier" way of binding global events to actions within a component? If not, this is fine. But it feels like Vue probably has a way to do this that might be a little less bare metal than using `addEventListener`.

Also, there's a nice-loooking library called [mousetrap](https://craig.is/killing/mice) that you might want to try out here. (You might not, seeing as you have a nice short implementation for keybindings here already, but just passing it along.)

waxlamp (dismissed):

src/components/Controls.vue
JackWilb: It's easy to bind key combos to certain DOM elements, but globally, it's more tricky. Here are a few things I found:
- [How to capture keystrokes, vue forum](https://forum.vuejs.org/t/capture-keypress-for-all-keys/14560)
- [Using raw js instead of vue-shortcut](https://shubhamjain.co/til/vue-shortcuts/)
- [Using ctrl-v in an app, bind to 1 input](https://stackoverflow.com/questions/42711537/vuejs-how-to-detect-ctrlv)
    - > <input @keyup.ctrl.76="YourFn">

Mousetrap could be cool, but I think it adds a unnecessary dependency and we have a good implementation here already

waxlamp: Thanks for looking into it. What you've got looks like the best way to do it.

PR #99, opened 2020-08-13
Update dependencies, security vulns

JackWilb:


waxlamp (approved):
Was this a change that dependabot couldn't make on its own?

JackWilb:
> Was this a change that dependabot couldn't make on its own?

I think when the vuln was released to the world, the dependency versions had not been bumped so dependabot failed. I could have tried to do each one manually again, but I figured I would just clean it up manually

PR #106, opened 2020-08-20
Add alerts for missing graph and data too large

### LINKED ISSUE #88: Add notification that data is too large
Ultimately, we want to allow the user to filter the data when we detect that it's too large. This is being implemented in the api, so this more complicated step can likely be added soon (we could also redirect to the aql page that Jake is building as an intermediate step)

JackWilb:
Closes #88

For testing, I was using the `Boston` network (since it's larger than currently allowed). You can use any made up name for a missing network.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -4,6 +4,8 @@ command = "yarn build"
 
 [context.production.environment]
 VUE_APP_MULTINET_HOST = "https://api.multinet.app/api"
+VUE_APP_MULTINET_CLIENT = "https://multinet.app/"
 
 [context.deploy-preview.environment]
 VUE_APP_MULTINET_HOST = "https://api.multinet.app/api"
+VUE_APP_MULTINET_CLIENT = "https://multinet.app/"
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -40,8 +40,8 @@ export default {
       widthVariables: [],
       colorVariables: [],
       linkWidthScale: scaleLinear().domain([0, 10]).range([2, 20]),
-      dataTooLarge: false,
-      graphDoesntExist: false,
+      loadError: false,
+      loadErrorData: {},
     };
   },
 
@@ -74,30 +74,39 @@ export default {
         return [];
       }
     },
-    client_link() {
-      return process.env.VUE_APP_MULTINET_CLIENT;
     },
-    aql_link() {
-      return `${process.env.VUE_APP_MULTINET_CLIENT}/workspaces/connectivity_matrix/aql`
-    },
-  },
 
   async mounted() {
     const { workspace, graph, host } = getUrlVars();
     if (!workspace || !graph) {
-      throw new Error(
-        `Workspace and graph must be set! workspace=${workspace} graph=${graph}`,
-      );
+      this.loadError = true;
+      this.loadErrorData.message = 'Workspace and graph must be set in the url.';
+      this.loadErrorData.buttonText = 'Back to Multinet';
+      this.loadErrorData.href = process.env.VUE_APP_MULTINET_CLIENT;
+      throw new TypeError('Workspace and graph must be set in the url.');
     }
     try {
       this.graphStructure = await loadData(workspace, graph, host);
     } catch (error) {
+      this.loadError = true;
+
+      // Set error message, button text, and href based on error type
       if (error instanceof DataTooBigError) {
+        this.loadErrorData.message = 'Your data is too large to view with this visualization. Please use AQL to reduce the size before you visualize it.';
+        this.loadErrorData.buttonText = 'AQL wizard';
+        this.loadErrorData.href = `${process.env.VUE_APP_MULTINET_CLIENT}/#/workspaces/${workspace}/aql`;
+      } else if (error.status === 404) {
+        this.loadErrorData.message = `Network ${this.graph} does not exist.`;
+        this.loadErrorData.buttonText = 'Back to multinet';
+        this.loadErrorData.href = process.env.VUE_APP_MULTINET_CLIENT;
+      } else {
+        this.loadErrorData.message = 'There has been an unexpected error.';
+        this.loadErrorData.buttonText = 'Back to Multinet';
+        this.loadErrorData.href = process.env.VUE_APP_MULTINET_CLIENT;
       }
 
-      if (error.status === 404){
-        this.graphDoesntExist = true;
-      }
+      // Re-throw the error from loadData
+      throw error;
     }
     
     this.provenance = setUpProvenance(this.graphStructure);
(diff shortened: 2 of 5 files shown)


src/components/Controls.vue
```diff
@@ -39,6 +39,8 @@ export default {
       widthVariables: [],
       colorVariables: [],
       linkWidthScale: scaleLinear().domain([0, 10]).range([2, 20]),
+      dataTooLarge: false,
+      graphDoesntExist: false,
```
waxlamp: We should prefer to avoid this sort of double-negative setup. That is, `graphMissing` (keeping the default `false`) or `graphExists` (flipping polarity to `true`) would both be preferable to a variable name that includes the word `not`.

src/components/Controls.vue
```diff
@@ -71,6 +73,12 @@ export default {
         return [];
       }
     },
+    client_link() {
+      return process.env.VUE_APP_MULTINET_CLIENT;
+    },
+    aql_link() {
+      return `${process.env.VUE_APP_MULTINET_CLIENT}/workspaces/connectivity_matrix/aql`
```
waxlamp: Why is `connectivity_matrix` hardcoded here?

src/components/Controls.vue
```diff
@@ -80,7 +88,19 @@ export default {
...
-    this.graphStructure = await loadData(workspace, graph, host);
+    try {
+      this.graphStructure = await loadData(workspace, graph, host);
+    } catch (error) {
+      console.error(error)
+      if (error instanceof RangeError) {
+        this.dataTooLarge = true;
+      }
+
+      if (error.status === 404){
```
waxlamp: This can probably be `else-if`ed into the previous clause.

src/components/Controls.vue
```diff
@@ -80,7 +88,19 @@ export default {
...
+        this.graphDoesntExist = true;
+      }
```
waxlamp: You may want to add an `else` clause to put up a generic error message banner.

src/components/Controls.vue
```diff
@@ -248,6 +268,31 @@ export default {
       <!-- node-link component -->
       <v-col>
         <v-row row wrap class="ma-0 pa-0">
+          <v-alert 
+            type="error" 
+            :value="dataTooLarge"
+            prominent
+          >
+            <v-row align="center">
+              <v-col class="grow">Your data is too large to view with this visualization. Please use AQL to reduce the size before you visualize it.</v-col>
```
waxlamp: We will want a better guidance here eventually. "Use AQL to fix it" will be alienating to newbies. Could you file an issue about this?

netlify.toml
```diff
@@ -4,6 +4,8 @@ command = "yarn build"
 
 [context.production.environment]
 VUE_APP_MULTINET_HOST = "https://api.multinet.app/api"
+VUE_APP_MULTINET_CLIENT = "https://multinet.app/"
 
 [context.deploy-preview.environment]
 VUE_APP_MULTINET_HOST = "https://api.multinet.app/api"
+VUE_APP_MULTINET_CLIENT = "https://multinet.app/"
```
jjnesbitt, line 11: FYI since these two contexts share the same exact configurations, these variable declarations can instead be put under the `[build.environment]` section. I know it's not something for this PR, but this is the first time I've seen it so I thought I'd mention it.

src/components/Controls.vue
```diff
@@ -248,6 +268,31 @@ export default {
...
+              </v-col>
+            </v-row>
+          </v-alert>
+
+          <v-alert 
+            type="error"
+            :value="graphDoesntExist"
+            prominent>
+            <v-row align="center">
+              <v-col class="grow">It seems like that graph doesn't exist.</v-col>
```
waxlamp: Don't hedge here: instead say something like `Network "${networkName}" doesn't exist`.

src/lib/multinet.ts
```diff
@@ -10,7 +10,7 @@ async function _downloadAllRows(api: any, workspace: string, tableName: string,
     (table.count > 100 && tableType === 'node') ||
     (table.count > 2000 && tableType === 'link')
   ) {
-    throw new Error(`The table called ${tableName} is too large, not downloading.`);
+    throw new RangeError(`The table called ${tableName} is too large, not downloading.`);
```
waxlamp: I don't think this is an appropriate use of `RangeError`, going by the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError).

You can instead just define a class `DataTooBig` or something like that.

src/components/Controls.vue
```diff
@@ -248,6 +268,31 @@ export default {
...
+              <v-col class="shrink">
+                <v-btn :href="this.client_link">Back to multinet</v-btn>
+              </v-col>
+            </v-row>
+          </v-alert>
```
waxlamp: A lot of code repetition here. Seems it would be better to set the content of the alert in a data prop, then use `v-if` to control whether it appears or not. That way you could support other error messages in the future without copy-pasting this again.

src/components/Controls.vue
waxlamp: I'm also not sure that redirecting to the Multinet client is the best course of action here. After this page is replaced by the client, and they do something with AQL, should they then reopen the view? That will result in two instances of the Multinet client (the original one, plus this redirected one) and a new adjacency matrix view.

It is ok to leave for now but I'd like to have the discussion about how to handle this in general.

waxlamp (changes requested):
The visual effect is good, but I identified a few code changes we should make before merging.

src/lib/multinet.ts
JackWilb: Made a custom error class. See 75edce8

src/components/Controls.vue
JackWilb: Agreed, I removed this by refactoring the logic into 1 bool and 1 error object that can have multiple messages in 
0e8c999

src/components/Controls.vue
JackWilb: Typo. Fixed in 0e8c999

src/components/Controls.vue
JackWilb: Fixed in 0e8c999

src/components/Controls.vue
JackWilb: Fixed in 0e8c999

src/components/Controls.vue
JackWilb: Fixed in 0e8c999

src/components/Controls.vue
JackWilb: I removed this in 0e8c999, but there is still a lot of replication when setting `this.loadErrorData`. Let me know what you think to my solution

src/components/Controls.vue
JackWilb: Opened #107

netlify.toml
JackWilb: How do you feel about this change @waxlamp? I think being verbose about which variables apply in which contexts is useful, especially if we plan to have different endpoints for some (e.g. `VUE_APP_MULTINET_CLIENT = "https://testing.multinet.app/"` for deploy previews). I also see a benefit to not being redundant.

jtomeck (approved):

netlify.toml
waxlamp: I think it's ok to have it split up like this, since it makes it more explicit that these *might* vary in the future.

For instance, shouldn't the `HOST` variable be pointing to a testing API?

src/components/Controls.vue
waxlamp: Looks much better :+1:

src/components/Controls.vue
waxlamp: Good change.

It seems to me that you don't really even need the `bool`--`loadError` could instead be a computed prop that checks whether `loadErrorData.message` is empty or not. But I'll leave that change to your discretion.

src/components/Controls.vue
waxlamp: Thanks--added a note to that issue discussing the idea of doing more to support the user in an error condition.

waxlamp (approved):

PR #112, opened 2020-09-03
Add prettier to the codebase and format code

### LINKED ISSUE #110: Add prettier for code formatting
Replicate the changes from the matrix.

JackWilb:
Closes #110.

This PR makes `yarn lint` use the `--no-fix` option by default, adds prettier, adds the commands to run prettier (`format:check`, `format:fix`), formats the code, and adds the format check to the CI.


src/App.vue
```diff
@@ -25,7 +25,7 @@ export default {
 
 <style>
 #app {
-  font-family: "Avenir", Helvetica, Arial, sans-serif;
+  font-family: 'Avenir', Helvetica, Arial, sans-serif;
```
waxlamp, line 28: This should be double quotes, right?

I saw that you specified a double-quotes rule for CSS files; any idea what's going on here?

src/components/NodeLink/functionUpdateVis.ts
```diff
@@ -371,28 +431,45 @@ export function drawNested(
...
+          `${((nodeMarkerHeight - 16 - 5 - 5) * d[barVar]) / maxValue}px`,
+      )
+      .attr(
+        'y',
+        (d: Node) =>
+          `${
+            nodeMarkerHeight -
+            5 -
+            ((nodeMarkerHeight - 16 - 5 - 5) * d[barVar]) / maxValue
+          }px`,
```
waxlamp, line 470: When you see the formatter making changes such as this, it's a good clue that some sort of refactoring is in order.

In this case, I'd recommend defining a `const` variable to compute that expression with the `- 16 - 5 - 5` part so that the final value of the function can be expressed more concisely. That sequence of subtractions is so odd that I would include a comment explaining it. And finally, the two instances of that expression occur in functions that are almost identical--maybe there's something that can be done there as well (maybe not).

Just to be clear, this is an example of the sort of general advice I'd give upon seeing weird formatting like this. Don't make those recommended changes on this branch (since this branch is just meant for getting the prettier infrastructure into place).

waxlamp (changes requested):
Looking good!

There seems to be a problem with quotation marks in CSS, but otherwise this looks ready to go.

JackWilb:
Closing in favor of adding elint-formatter

PR #114, opened 2020-09-30
Add eslint-formatter and fix all linter issues

JackWilb:
This depends on #113.

Adds eslint-formatter, and runs the lint:fix command for all auto fixable problems. I then manually cleaned up the remaining issues


.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    '@vue/typescript/recommended',
+  ],
+
+  plugins: [
+    'vuetify',
+  ],
+
+  rules: {
+    'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
```
waxlamp: Maybe these should raise errors instead of warnings. I can't remember if we treat warnings as errors in our build step.

.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    'max-len': ['off'],
```
waxlamp, line 23: Does the formatter attempt to break long lines? If it does, I would set this to something like 100 so that the formatter at least has a chance to simplify our code. I like the situation where, if the formatter can only break up your long line by making it harder to read, it's a signal to you to refactor a bit for readability. If you allow unlimited line length, you lose this benefit.

.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    'import/prefer-default-export': ['off'],
+    'no-underscore-dangle': ['off'],
```
waxlamp, line 26: Personally, I think this is a good rule to leave as a default. Underscores just aren't needed to indicate privateness, etc., since JavaScript/TypeScript have good export mechanisms for "publishing" functions from a module.

.eslintrc.js
```diff
@@ -0,0 +1,36 @@
...
+    'no-param-reassign': ['error', { props: false }],
```
waxlamp, line 28: Do you need this line? It seems like this is the default value for `no-param-reassign`.

src/components/NodeLink/Legend.vue
```diff
@@ -1,4 +1,5 @@
 <script>
+/* eslint-disable vue/no-mutating-props */
```
waxlamp, line 2: Why do this? Mutation of props is fundamentally a bad idea in Vue.

waxlamp (commented):
Looks good overall. I left several questions/suggestions that are meant to promote a bit of discussion (important since we'll likely use whatever we come up with here for the other vis apps we develop).

src/components/NodeLink/Legend.vue
JackWilb: It's the way the app works currently, and I didn't want to make a big modification other than linter changes in this PR. We need to refactor it so we're using best practices with emits

.eslintrc.js
JackWilb: Yes, without it there are a lot of errors. I'm not sure if it's default in a different version of eslint, and not ours

.eslintrc.js
JackWilb: I'm mostly using it for unused params in arrow functions, not just privateness. You still think that we should remove it?

.eslintrc.js
JackWilb: I think this was in the default .eslintrc.js. It's also in our mutlinet-client repo so keeping it allows us to maintain parity with that config.

.eslintrc.js
JackWilb: Agreed. I've updated it in cc3210b

src/components/NodeLink/Legend.vue
waxlamp: Got it. Can you file an issue to fix this? It's something we can discuss at the rearchitecting meeting too.

.eslintrc.js
waxlamp: Right, but we don't need to necessarily keep parity with what's in multinet-client currently; our goal should be to have the best possible eslint setup and use it in all our projects.

So we can leave this as-is for now, but please file an issue to discuss whether we want line-length limits.

.eslintrc.js
waxlamp: If you don't use underscore for unused parameters in an arrow function, does some other rule get triggered? I don't think you need the underscore style for that purpose; if an arrow function is using the second argument but not the first, I think you should just name the first argument normally and then not use it (my gut tells me eslint should allow that, since it's a common situation when passing arrow functions to a framework).

.eslintrc.js
waxlamp: Which errors do you man? This rule seems to say that it will *cause* errors if you are reassigning parameters or their properties.

JackWilb: You said this was the default, but when I remove this line, I get 25 errors like this:

```
error: Assignment to property of function parameter 'd' (no-param-reassign) at src/lib/multinet.ts:73:51:
  71 | // eslint-disable-next-line  @typescript-eslint/no-explicit-any
  72 | function _defineNeighbors(nodes: any[], links: any[]) {
> 73 |   nodes.forEach((d: { neighbors: string[] }) => { d.neighbors = []; });
     |                                                   ^
  74 | 
  75 |   links.forEach((link) => {
  76 |     nodes.filter((d: Node) => d._id === link._from)[0].neighbors.push(link._to);
```

.eslintrc.js
JackWilb: Looks like my memory was wrong, it was for handling multinet columns from Arango, such as _key, _from, _to, etc. Here's an example of the warning that is thrown without disabling this rule:

```
error: Unexpected dangling '_' in '_from' (no-underscore-dangle) at src/lib/multinet.ts:77:69:
  75 |   links.forEach((link) => {
  76 |     nodes.filter((d: Node) => d._id === link._from)[0].neighbors.push(link._to);
> 77 |     nodes.filter((d: Node) => d._id === link._to)[0].neighbors.push(link._from);
     |                                                                     ^
  78 |   });
  79 |   return nodes;
  80 | }
```

src/components/NodeLink/Legend.vue
JackWilb: Opened #116

.eslintrc.js
waxlamp: Ah yeah, this is annoying. I think your rule is fine, though I guess I'd toy with using the subscripting form for these uses (e.g., `link['_to']'`etc. If you keep your eslint rule, I'd just request that you leave a comment explaining why it's there.

.eslintrc.js
JackWilb: Opened #117

.eslintrc.js
waxlamp: Ok I see. Same as above, I'd put in a comment explaining that you need this rule override in order to allow D3-style mutating callbacks.

As a general point, disabling a rule just to suppress errors (with no further justification) is in general a no-go. In this case, by contrast, we do have a good reason to allow parameter property mutation--namely that it's a very common pattern for D3. Even then, I'd prefer in general to use inline disable directives to accommodate things like D3 (so that in non-D3 contexts you aren't doing this), but in this case I will leave it up to you. As above, I just would request a comment in the eslint rules explaining why we're doing it (in particular, so that we don't blindly copy this rule to other projects where it may not apply).

JackWilb: Fixed in 831e857 by adding a comment

.eslintrc.js
JackWilb: Fixed in 831e857 by adding a comment

waxlamp (approved):
Looks good, @JackWilb, thanks!

PR #121, opened 2020-11-16
Add direct-vuex store and move rendering logic into vue component

### LINKED ISSUE #71: Use d3 scales for the bars

### LINKED ISSUE #101: Marker size control doesn't have any effect anymore
after the recent PR for issue #87, the marker resize slider doesn't have any effect anymore.

### LINKED ISSUE #108: Add more type safety through the app
Currently the vue components are not using typescript in the <script> sections.

JackWilb:
I think this was fixed before, but we should definitely close #71 since we're using scaleLinear for the bars in the re-write
Closes #101 
Closes #108

This is the first of a few steps to get the app to a state where it's easier to get up to speed with the app, and to add new features.

In this gargantuan change set you will find the following:
- The addition of a VueX store
- Migration of logic from separate ts files into the components (no class to worry about with the nodelink)
- The reduction of D3 logic where possible, in favor of using Vue for as much of the DOM rendering as possible
- Removal of the provenance library (to be added back by tracking the VueX store in a follow up)
- Migration of all components to TypeScript
- All components now use Vue.extend() for better TypeScript typing
- Integration of CSS into the components
- All functionality is preserved

TODOs for (hopefully!) smaller follow ups:
- [x] File issue set reactivity, performance, new Set(). (#127)
- [x] File issue for setting host from querystring again. (#126)
- [x] File issue for using composition API (#125)
- [x] ~~File issue for re-implementing provenance~~ See #122 for new implementation of the provenance library
- [x] `simulationLinks` should be made from a deep copy of network so the original network is not mutated (this will make the download button work much better) (#130)
- [x] The bar charts in the legend need to be modified to order the elements (currently makes the width scale not work exactly as intended) (#132)
- [x] Clean out dead CSS (#131)
- [x] Add controls for allowing directional edges (#129)
- [x] ~~Nodes can wander off the svg~~ See #123
- [x] Error message always seems to be "Unexpected Error Occurred", we should have better ones than that (#128)


waxlamp (approved):
I've looked this change set over but I don't think there's any meaningful way I can provide an actual review. In lieu of that, I'm rubber stamping this with the understanding that if anything breaks we'll fix it in small, targeted PRs (as we usually do, excluding this one).

Thanks for all the effort on this, @JackWilb.

JackWilb:
Sounds good. Thanks, Roni.

PR #122, opened 2020-12-07
Add provenance tracking to select and de-select nodes

JackWilb:
Depends on #121, I'll retarget this branch to master once it's merged (this just helps with seeing the changes).

Adds the most recent version of the provenance library that I've been told "will be released without changes to the API". That means the logic won't have to change for us to use the 2.0.0 release of the library.

I think the provenance library cannot track objects that are vue reactive, because it will get into an infinite loop. To get around that problem, I read the state in with `JSON.parse(JSON.stringify(state))` to get rid of the reactivity and then I make sure when I update, I don't pass a reactive object. Kiran told me "Deep-diff and trrack both work with only pure JSON objects... I added serialization support for Sets", so passing a set through as I have done, should be okay.

You'll find that the first thing that I added to the provenance is selecting and deselecting nodes. I figured I'd keep this change small, and I'll address the other pieces that we would want to update with provenance in a follow up. We can also implement a provenance visualization that has been created by our team at Utah that works on Trrack.

We can also support loading an entire state from a provenance querystring, if we'd like. So that's something to consider for additional follow ups.

### DIFF
--- a/src/lib/provenanceUtils.ts
+++ b/src/lib/provenanceUtils.ts
@@ -5,6 +5,7 @@ export function updateProvenanceState(vuexState: State, label: ProvenanceEventTy
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const stateUpdateActions = createAction<State, any[], ProvenanceEventTypes>((provState, newProvState) => {
     if (label === 'Select Node' || label === 'De-select Node') {
+      // eslint-disable-next-line no-param-reassign
       provState.selectedNodes = newProvState.selectedNodes;
     }
   })
--- a/src/lib/provenanceUtils.ts
+++ b/src/lib/provenanceUtils.ts
@@ -1,0 +1,16 @@
+import { ProvenanceEventTypes, State } from '@/types';
+import { createAction } from '@visdesignlab/trrack';
+
+export function updateProvenanceState(vuexState: State, label: ProvenanceEventTypes) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const stateUpdateActions = createAction<State, any[], ProvenanceEventTypes>((provState, newProvState) => {
+    if (label === 'Select Node' || label === 'De-select Node') {
+      provState.selectedNodes = newProvState.selectedNodes;
+    }
+  })
+    .setLabel(label);
+
+  if (vuexState.provenance !== null) {
+    vuexState.provenance.apply(stateUpdateActions(vuexState));
+  }
+}
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -13,6 +13,7 @@ import {
 } from 'd3-scale';
 import { schemeCategory10 } from 'd3-scale-chromatic';
 import { initProvenance } from '@visdesignlab/trrack';
+import { updateProvenanceState } from '@/lib/provenanceUtils';
 
 Vue.use(Vuex);
 
@@ -167,11 +168,19 @@ const {
 
     addSelectedNode(state, nodeID: string) {
       state.selectedNodes = new Set(state.selectedNodes.add(nodeID));
+
+      if (state.provenance !== null) {
+        updateProvenanceState(state, 'Select Node');
+      }
     },
 
     removeSelectedNode(state, nodeID: string) {
       state.selectedNodes.delete(nodeID);
       state.selectedNodes = new Set(state.selectedNodes);
+
+      if (state.provenance !== null) {
+        updateProvenanceState(state, 'De-select Node');
+      }
     },
 
     setRenderNested(state, renderNested: boolean) {
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -4,56 +4,16 @@ import { createDirectStore } from 'direct-vuex';
 import { Simulation } from 'd3-force';
 
 import {
-  Link, Node, Network, SimulationLink,
+  Link, Node, Network, SimulationLink, State, LinkStyleVariables, LoadError, NestedVariables,
 } from '@/types';
 import api from '@/api';
 import { GraphSpec, RowsSpec, TableRow } from 'multinet';
 import {
-  scaleLinear, ScaleLinear, scaleOrdinal, ScaleOrdinal,
+  scaleLinear, scaleOrdinal,
 } from 'd3-scale';
 import { schemeCategory10 } from 'd3-scale-chromatic';
 
 Vue.use(Vuex);
-
-interface LoadError {
-  message: string;
-  buttonText: string;
-  href: string;
-}
-
-interface NestedVariables {
-  bar: string[];
-  glyph: string[];
-}
-
-interface LinkStyleVariables {
-  width: string;
-  color: string;
-}
-
-interface AttributeRanges {
-  [key: string]: {attr: string; min: number; max: number};
-}
-
-export interface State {
-  workspaceName: string | null;
-  networkName: string | null;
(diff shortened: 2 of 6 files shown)


JackWilb:
This is waiting on an update to the Trrack library that will add types for deep-diff. We're the first typescript project to use `"strict": true` in the typescript compiler, and thus disallow implicit anys, so the first project to catch this issue.

src/lib/provenanceUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (label === 'Select Node' || label === 'De-select Node') {
+      provState.selectedNodes = newProvState.selectedNodes;
+    }
+  })
+    .setLabel(label);
+
+  if (vuexState.provenance !== null) {
+    vuexState.provenance.apply(stateUpdateActions(vuexState));
+  }
+}
```
waxlamp: Can you help me understand what this function is doing?

src/store/index.ts
```diff
@@ -4,57 +4,19 @@ import { createDirectStore } from 'direct-vuex';
...
-  fontSize: number;
-  labelVariable: string;
-  colorVariable: string;
-  selectNeighbors: boolean;
-  nestedVariables: NestedVariables;
-  linkVariables: LinkStyleVariables;
-  attributeRanges: AttributeRanges;
-  nodeColorScale: ScaleOrdinal<string, string>;
-  linkWidthScale: ScaleLinear<number, number>;
-}
```
waxlamp, line 56: Why move this stuff out? They're types intimately connected to state tracking, so aren't really generally consumable types outside of the project.

Not saying you can't move these to a top level types.ts file, but I was wondering about your reasoning.

src/store/index.ts
```diff
@@ -205,11 +168,19 @@ const {
 
     addSelectedNode(state, nodeID: string) {
       state.selectedNodes = new Set(state.selectedNodes.add(nodeID));
+
+      if (state.provenance !== null) {
+        updateProvenanceState(state, 'Select Node');
+      }
```
waxlamp, line 174: This is nice and compact, but something about it feels funny. The VueX state is a reactive object, so you could create something like our multimatrix "event bus" to respond to state changes with its own actions, such as these side-effectful updates of the Trrack state. You could even equip that "bus" with computed props that would provide JSON representations of bits of the VueX state and conceptually streamline the whole thing.

If you decide on that, it can be done in a followup. But since you're testing out an approach here for the first time, I thought I'd mention it.

src/store/index.ts
```diff
@@ -260,6 +231,14 @@ const {
         state.linkWidthScale.domain(domain).range([1, 20]);
       }
     },
+
+    createProvenance(state) {
+      state.provenance = initProvenance<State, ProvenanceEventTypes, unknown>(
```
waxlamp, line 238: This goes somewhat along with my last comment: Since the provenance is for an independent system that reacts to changes in the state (and possibly causes some changes of its own) from an engineering perspective it seems like it should be encapsulated *outside* of the VueX state.

src/store/index.ts
JackWilb: Having the types here creates a cyclic import between the provenance library and the vuex store. We need the types in the provenance and we need the provenance in the store. I figured the easiest resolution was to just move the types to types, but if you have some other ideas, I'm open to them

waxlamp: Oh ok. Yeah cyclic dependencies are a pain. You're right, the best (only?) thing to do is to factor out the cyclically dependent pieces into a common location.

src/lib/provenanceUtils.ts
JackWilb: Yes, it takes in the new state (`vuexState`) and a label for the type of event, generates what the action function should be (in this case, just updating the selected nodes in the provenance state with the selected nodes from the new state), and applies that onto the provenance.

The utility of this action function syntax is that you can do more than just update the store. In other applications, here is where you'd do a d3 update, or something similar. Since our app has all the DOM logic handled by Vue, we just wanted to track the state changes, not apply any additional logic.

src/store/index.ts
JackWilb: I don't really see how that implementation would be much different from what we already have, but I could just not be understanding what you're saying. It sounds to me like you'd like me to replace this code chunk with a similar code chunk, something like

```
eventBus.updateProvenance()
```

to update the state whenever something changes. If that's the case, I don't think there's much utility. If I'm wrong, would you provide me a bit of pseudo-code so I can see what you mean?

waxlamp: Well the idea would be to just let the VueX store manage state, and rely on *other* components to receive updates and perform these side effects.

If you look at any of the components that currently use VueX state, you'll see that you're doing stuff like creating computed props based on state values, then reacting to those locally. What I'm suggesting is to create a controller component that behaves similarly, that would wrap Trrack and watch the VueX values in order to trigger these sorts of actions. Since it's not a "Vue component" in the sense of rendering stuff to the DOM, you could create such a thing almost exactly the way we created that stopgap event bus component.

So, just to summarize, it's not a matter of replacing the code you have with something like `eventBus.updateProvenance()`; it would be closer to supplying the VueX store to `eventBus` when you construct it, then having computed props in that component, and using something like `watch()` or `watchEffect()` in that component to respond to VueX state changes. The major upshot here is reusability: if my design idea actually works, you'll have a fairly general component that can be used in any application that uses both Trrack and VueX in order to add provenance to any Vue app.

What you've got here is simple enough to justify including as-is, but this larger design idea I'm describing would scale better (in case you had other libraries you wanted to bring in in a similar way) and it could be replicated in a modular way to use elsewhere. As I mentioned above, we could roll with this for now and revisit this design in a followup if you feel its sufficiently valuable.

I hope this made sense. I drank two cups of coffee today instead of my usual single cup, and my wife says I am more chatty than usual :upside_down_face:

JackWilb: Thanks for the ideas, Roni! It does sound like this would be something worth exploring in a follow up PR. I'm curious how we could initialize the "eventBus" component, since we can't use a traditional vue component, like you said. I guess we could just mount it somewhere like App.vue (the highest level component that mounts everything else) and then pass the VueX store. On the whole, though, it does seem like this would be useful in the case where we have more libraries like this.

src/store/index.ts
JackWilb: If we follow your suggestion from above, we can have this be a method in the provenance "component" and then call it from wherever we bind data to the provenance (I suggested App.vue above).

For now, I can move the initialization out to the provenanceUtils.ts file, if that sounds like a reasonable compromise

src/store/index.ts
waxlamp: I think with composition API, you can just declare reactive data anywhere, and it will "just work". I'm not super sure about that though. Jake may have a better understanding.

src/store/index.ts
waxlamp: You can move this function if you want, but it's not necessary right now. If you want to file an issue to re-examine the design later, that would be more than enough to get this PR out the door.

Let me know what you'd like to do here, and then I can resolve this thread.

waxlamp (dismissed):
Pre-approving even though there appears to be a breakage caused by `trrack`. Once that's fixed I can re-approve.

src/store/index.ts
JackWilb: Opened #137

waxlamp (approved):

PR #124, opened 2020-12-07
Enable eslint rules that were disabled before refactor

### LINKED ISSUE #116: Refactor the modification of props to emit events
Currently we have a couple instances where we modify the props instead of emitting an event to the parent. This is not best practices and opens us up to bugs.

### LINKED ISSUE #117: Decide whether we should enforce line length limits
[Comment](https://github.com/multinet-app/view-nodelink/pull/114#discussion_r513559773)

JackWilb:
Depends on #121 

Closes #117 when we address my line length question
The error thrown for #116 is no longer an issue after #121. This removes the no-mutating-props exception, thus closes #116 

I've re-enabled `no-mutating-props` and `no-param-reassign` app wide, and have disabled the rule on the specific lines that need it. I also modified the `no-underscore-dangle` to explicitly allow `_id`, `_from`, and `_to`, since those are the actual values we need from arango, as opposed to disabling the rule entirely.

Also, we have no line length limit at the moment and it's not causing any issues. I think enforcing a line length limit just adds development overhead for refactoring long lines, without much added readability. I'd love to get your thoughts, though. Leaving no line length limits means that our configuration closely matches the same from the [multinet-client](https://github.com/multinet-app/multinet-client/blob/master/.eslintrc.js).

### DIFF
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -36,10 +36,8 @@ export default Vue.extend({
     simulationLinks(): SimulationLink[] | null {
       if (this.network !== null) {
         return this.network.edges.map((link: Link) => {
-          const newLink = link;
-          newLink.source = link._from;
-          newLink.target = link._to;
-          return (newLink as SimulationLink);
+          const newLink: SimulationLink = { ...link, source: link._from, target: link._to };
+          return newLink;
         });
       }
       return null;
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -36,9 +36,10 @@ export default Vue.extend({
     simulationLinks(): SimulationLink[] | null {
       if (this.network !== null) {
         return this.network.edges.map((link: Link) => {
-          link.source = link._from;
-          link.target = link._to;
-          return (link as SimulationLink);
+          const newLink = link;
+          newLink.source = link._from;
+          newLink.target = link._to;
+          return (newLink as SimulationLink);
         });
       }
       return null;
@@ -197,7 +198,9 @@ export default Vue.extend({
       nodes.forEach((node) => {
         // If the position is not defined for x or y, generate it
         if (node.x === undefined || node.y === undefined) {
+          // eslint-disable-next-line no-param-reassign
           node.x = Math.random() * this.svgDimensions.width;
+          // eslint-disable-next-line no-param-reassign
           node.y = Math.random() * this.svgDimensions.height;
         }
       });
@@ -215,7 +218,9 @@ export default Vue.extend({
       event.preventDefault();
 
       const moveFn = (evt: Event) => {
+        // eslint-disable-next-line no-param-reassign
         node.x = (evt as MouseEvent).clientX - this.svgOffset.x - (this.markerSize / 2);
+        // eslint-disable-next-line no-param-reassign
         node.y = (evt as MouseEvent).clientY - this.svgOffset.y - (this.markerSize / 2);
         this.$forceUpdate();
       };
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -237,13 +237,14 @@ const {
     },
 
     setNestedVariables(state, nestedVariables: NestedVariables) {
-      const newNestedVars = nestedVariables;
-      // Remove duplicates from the nested variables
-      newNestedVars.bar = [...new Set(nestedVariables.bar)];
-      newNestedVars.glyph = [...new Set(nestedVariables.glyph)];
+      const newNestedVars = {
+        ...nestedVariables,
+        bar: [...new Set(nestedVariables.bar)],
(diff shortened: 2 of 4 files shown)


src/components/MultiLink/MultiLink.vue
```diff
@@ -36,9 +36,10 @@ export default Vue.extend({
     simulationLinks(): SimulationLink[] | null {
       if (this.network !== null) {
         return this.network.edges.map((link: Link) => {
-          link.source = link._from;
-          link.target = link._to;
-          return (link as SimulationLink);
+          const newLink = link;
+          newLink.source = link._from;
+          newLink.target = link._to;
+          return (newLink as SimulationLink);
```
waxlamp: Instead of using `as`, I'd recommend typing `newLink` on line 39 directly.

While you're at it, I'd say you should define `newLink` in one go as an object literal rather than declaring it and using assignment on its fields (you could do this with the spread operator). This is mostly a style consideration, but I think the code would be clearer if you made this change.

src/components/MultiLink/MultiLink.vue
```diff
@@ -197,7 +198,9 @@ export default Vue.extend({
       nodes.forEach((node) => {
         // If the position is not defined for x or y, generate it
         if (node.x === undefined || node.y === undefined) {
+          // eslint-disable-next-line no-param-reassign
```
waxlamp, line 199: Are you certain this code triggers this error? Manipulating the elements of an array using `forEach()` is such a common pattern that I'm surprised TS doesn't allow it by default, especially when you're not, strictly speaking, reassigning the parameter here, you're mutating it by assigning to one of its members.

src/store/index.ts
```diff
@@ -237,14 +237,15 @@ const {
...
+      const newNestedVars = nestedVariables;
       // Remove duplicates from the nested variables
-      nestedVariables.bar = [...new Set(nestedVariables.bar)];
-      nestedVariables.glyph = [...new Set(nestedVariables.glyph)];
+      newNestedVars.bar = [...new Set(nestedVariables.bar)];
+      newNestedVars.glyph = [...new Set(nestedVariables.glyph)];
 
       // Allow only 2 variables for the glyphs
-      nestedVariables.glyph.length = nestedVariables.glyph.length > 2 ? 2 : nestedVariables.glyph.length;
+      newNestedVars.glyph.length = newNestedVars.glyph.length > 2 ? 2 : newNestedVars.glyph.length;
```
waxlamp: I think  you could clarify this line by using `Math.max()` instead of the ternary operator.

src/store/index.ts
```diff
@@ -237,14 +237,15 @@ const {
     },
 
     setNestedVariables(state, nestedVariables: NestedVariables) {
+      const newNestedVars = nestedVariables;
       // Remove duplicates from the nested variables
-      nestedVariables.bar = [...new Set(nestedVariables.bar)];
-      nestedVariables.glyph = [...new Set(nestedVariables.glyph)];
+      newNestedVars.bar = [...new Set(nestedVariables.bar)];
+      newNestedVars.glyph = [...new Set(nestedVariables.glyph)];
```
waxlamp: I'd recommend defining this with an object literal (similarly to my suggestion above).

src/store/index.ts
```diff
@@ -322,7 +323,9 @@ const {
 
       if (context.state.network !== null) {
         context.state.network.nodes.forEach((n: Node) => {
+          // eslint-disable-next-line no-param-reassign
```
waxlamp, line 327: Same question here as above.

waxlamp (dismissed):
Looks good! Approving despite leaving my suggestions, since I'll leave those to your discretion.

src/components/MultiLink/MultiLink.vue
JackWilb: Agreed, this is a much better solution. Typing it as `SimulationLink` requires that we use the object literal.

Fixed in 927704a

src/components/MultiLink/MultiLink.vue
JackWilb: Positive, here's the error:

`error  Assignment to property of function parameter 'node'  no-param-reassign`

src/store/index.ts
JackWilb: Same answer as above

src/store/index.ts
JackWilb: Fixed in e541f28

We actually needed min, not max

src/store/index.ts
JackWilb: Fixed in e541f28

JackWilb:
Thanks for the thorough review @waxlamp, I've addressed all your comments.

Had to force push to use min, not max

src/components/MultiLink/MultiLink.vue
waxlamp: Huh. Ok then. Carry on :cat:

waxlamp (approved):

Issue #131, opened 2020-12-08
Clean out some dead CSS


JackWilb:
When moving the CSS with the re-architect (#121), I just moved the code into the component and didn't remove any dead code. This should be a quick clean up

PR #133, opened 2020-12-09
Add directional edge markers and control toggle

### LINKED ISSUE #129: Implement directional edges toggle

JackWilb:
Depends on #122 (since the type definition moved to types.ts from store/index.ts)

Closes #129 
Is a first step towards #131 

This adds a text path element onto the link by using an href and then moves the arrow along the path using an offset. I also added another switch for the control in the control panel. There was a type that needed to be updated, `State`, and I enabled a new getter and setter in the store for the `directionalEdges` variable. I also cleaned out some old CSS that was no longer needed.

Here's the relevant MDN docs for [textPath](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/textPath) and here is the documentation for [startOffset](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/startOffset)

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -110,6 +110,15 @@ export default Vue.extend({
         store.commit.setSelectNeighbors(value);
       },
     },
+
+    directionalEdges: {
+      get() {
+        return store.getters.directionalEdges;
+      },
+      set(value: boolean) {
+        store.commit.setDirectionalEdges(value);
+      },
+    },
   },
 
   methods: {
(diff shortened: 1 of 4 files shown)


waxlamp (dismissed):
This is certainly a good start, but there are better (but more complicated) ways to indicate edge directionality with less clutter when lots of edges intersect.

Let's get some design feedback from Jared and Alex/Miriah.

JackWilb:
Force pushed because the branch this came from was interactive rebased. I had to remove some commits that were no longer in the base branch

waxlamp (approved):

PR #134, opened 2020-12-11
Make a deep copy of the network for simulationLinks

### LINKED ISSUE #130: Simulation links should be set from a deep copy so that the links in network are untouched
This will allow us to fix the downloads. It could introduce a bug when loading provenance, though

JackWilb:
Closes #130

I decided to use a `JSON.parse(JSON.stringify())` since the data that we get from MultiNet is made of properties that are all JSON serializable, and there's no easier option that doesn't require importing more libraries.

This means that the simulation doesn't update the `source` and `target` properties on the links in the store with references to the nodes that are the source and target.

### DIFF
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -36,7 +36,11 @@ export default Vue.extend({
     simulationLinks(): SimulationLink[] | null {
       if (this.network !== null) {
         return this.network.edges.map((link: Link) => {
-          const newLink: SimulationLink = { ...link, source: link._from, target: link._to };
+          const newLink: SimulationLink = {
+            ...JSON.parse(JSON.stringify(link)),
+            source: link._from,
+            target: link._to,
+          };
           return newLink;
         });
       }


JackWilb:
Rebased onto master and resolved the conflict caused by change from https://github.com/multinet-app/multilink/pull/124#discussion_r541280075.

src/components/MultiLink/MultiLink.vue
```diff
@@ -36,7 +36,11 @@ export default Vue.extend({
     simulationLinks(): SimulationLink[] | null {
       if (this.network !== null) {
         return this.network.edges.map((link: Link) => {
-          const newLink: SimulationLink = { ...link, source: link._from, target: link._to };
+          const newLink: SimulationLink = {
+            ...JSON.parse(JSON.stringify(link)),
+            source: link._from,
+            target: link._to,
+          };
```
waxlamp, line 43: I agree that going through JSON and back comprehensively deep-copies an object, but it's very slow. I'd open an issue to track any performance problems that might result from this.

waxlamp (approved):

src/components/MultiLink/MultiLink.vue
JackWilb: Opened #138

Issue #125, opened 2020-12-08
Use the composition API over the current options API


JackWilb:
The composition API will provide us much better code clarity, and means that we can combine all the logic and definitions together instead of in a fragmented way

PR #139, opened 2020-12-23
Add composition API and use it in App.vue

JackWilb:
A first step towards #125

This add the composition API as dependency, initializes it (With Vue.use()), and migrates App.vue to use it. I chose App.vue since it's the simplest component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "dependencies": {
     "@types/d3": "^5.7.2",
     "@visdesignlab/trrack": "^2.0.0-alpha.7",
+    "@vue/composition-api": "^1.0.0-beta.22",
     "core-js": "^3.4.3",
     "d3-array": "^2.8.0",
     "d3-axis": "^2.0.0",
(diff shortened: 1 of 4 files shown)


waxlamp (approved):
Looks great!

One small nit (completely at your discretion): I think the `Container` in `multilinkContainer` and `multilinkContainerDimensions` is superfluous. The type annotations and computed prop definitions express precisely what those are, and "container" is just a beast of a word to type out.

JackWilb:
> Looks great!
> 
> One small nit (completely at your discretion): I think the `Container` in `multilinkContainer` and `multilinkContainerDimensions` is superfluous. The type annotations and computed prop definitions express precisely what those are, and "container" is just a beast of a word to type out.

I actually think there is a distinction and we should keep `Container`. The ref refers to the container that holds the diagram -- it's on the v-col, not the MultiLink element -- so it's not superfluous. Also, this is the only component that references it for now, and if that changes, we can refactor the variable names to something smaller.

PR #140, opened 2021-01-05
Refactor sidebar to use the new style

### LINKED ISSUE #105: Copy sidebar styling from the adjmatrix
Once multinet-app/view-adjmatrix#82 lands, we can copy over the styling from there to make the app feel more cohesive.

JackWilb:
Closes #105

I refactored how the app mounts the components and how the sidebar looks. This uses the same style and structure as the MultiMatrix, with the addition of the legend panel below the controls. Overall, I love the look and I'm open to your feedback.

I think it currently looks a bit cluttered (too many buttons perhaps?). @jtomeck, would you mind taking a look and seeing if there's anything that might look a bit better?

TODO (likely for follow ups):
- [x] Fix the sizing of the legend drop targets (#142)


jtomeck (changes requested):
This looks really good. The only thing I'd like to see is better hierarchy of the action buttons (start, stop, release, export). To me, start and stop seem like the main simulation controls, while "release pinned nodes" seems like you're just clearing what you have selected. I think those 3 buttons should be paired together, with "export" as the main and final action below them.

I suggest we do the following:
- Simplify the wording of the buttons to "Start," "Stop," and "Release Pinned."
- Add play icon to start, stop icon to stop, and either `mdi-pin-off` or `mdi-debug-step-out` to release pinned.
- Right align "Start" and "Stop" side-by-side
- Left align "Release Pinned" on the other side of "Start" and "Stop" (this might require you to use a smaller size prop for the buttons)
- Keep "Start" and "Stop" the primary color with the `depressed` prop, but give "Release Pinned" the `text` prop and maybe make it `grey darken-3`
- Keep the export button below the 3 other buttons, full width, `depressed`, and `grey darken-3`

I think this would help the action buttons portion. Let me know what you think.

jtomeck (approved):

waxlamp (approved):
Looks good. We should explore whether to integrate the start and stop buttons into a single button that changes state when clicked, but this looks great at the moment. Thanks @JackWilb and @jtomeck!

JackWilb:
> Looks good. We should explore whether to integrate the start and stop buttons into a single button that changes state when clicked, but this looks great at the moment. Thanks @JackWilb and @jtomeck!

Good idea. I opened #141

PR #143, opened 2021-01-11
Use one button for simulation start/stop in the controls

### LINKED ISSUE #141: Consider using one button for animation start and stop
https://github.com/multinet-app/multilink/pull/140#pullrequestreview-562867800

JackWilb:
Closes #141

Reduces the visual clutter by making the start and stop buttons into just one button that changes based on the simulation status.

My initial route was to try and use a component level data variable, but that didn't work. When the simulation stopped naturally, it wouldn't update the button. The solution was to refactor the start stop logic to the store and to have the simulation tell the store when it had finished/started. Now the buttons are in sync.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -277,9 +277,8 @@ export default Vue.extend({
           />
 
           <v-row>
-            <v-col cols="5">
+            <v-col cols="6">
               <v-btn
-                class="px-2"
                 color="grey darken-3"
                 depressed
                 text
@@ -294,14 +293,13 @@ export default Vue.extend({
             </v-col>
 
             <v-col
-              cols="3"
-              class="px-0"
+              cols="6"
             >
               <v-btn
-                class="ml-2 px-1"
                 color="primary"
                 depressed
                 small
+                block
                 @click="stopSimulation"
               >
                 <v-icon small>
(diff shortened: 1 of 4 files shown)


jtomeck (dismissed):
@JackWilb I added some changes to clean up the spacing and placement of the icons. Let me know what you think.

src/components/Controls.vue
```diff
@@ -277,55 +281,58 @@ export default Vue.extend({
...
-                  mdi-stop
+                <v-icon
+                  left
+                  small
+                >
+                  mdi-play
                 </v-icon>
-                Stop
+                Start
               </v-btn>
```
waxlamp: If the only difference between these buttons are the text, icon, and what function is called, then I think you would be better off unifying this into a single button that uses a computed prop to fill in the icon name and button text, and create a function `toggleSimulation()` that checks the value of `simulationRunning` and runs either `stopSimulation()` or `startSimulation()`. That will halve this code, and make it more readable.

I think this would really improve the code, but I'll leave it to your discretion.

waxlamp (dismissed):
The only other comment I have about this is that the placement seems weird. The pin/unpin button appears right about the play/stop button, with dead space to the right of both buttons. Could we have this start/stop button in the same row as the pin/unpin button? Or do we want them this way for other organizational reasons?

Approving because what you've got here is good. Up to you whether to address my concerns now, later, or never.

src/components/Controls.vue
JackWilb: Fixed in d242a85

jtomeck (approved):

PR #144, opened 2021-01-11
Add search bar that selects node from autocomplete element

### LINKED ISSUE #48: Re-implement node searching
Allow searching for a node with a search box in the control panel

JackWilb:
Closes #48

Re-implements node searching by using an autocomplete vuetify element. The logic just takes the search term from the autocomplete and finds the right id to select in the nodes.

A possible logic issue we could bump into, is a node with the same key in 2 tables. I'm not sure how best to get around that, other than using id, which should be unique. That make the autocomplete really ugly and too wide though. I think this is a fine solution for now, but I'll open an issue to track this potential issue.

Here's a video of it in action:
https://user-images.githubusercontent.com/36867477/104361410-a483da00-54cf-11eb-8fc9-e2df2b577d14.mp4

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -3,11 +3,21 @@ import Vue from 'vue';
 import Legend from '@/components/MultiLink/Legend.vue';
 
 import store from '@/store';
-import { Node, Link } from '@/types';
+import { Node, Link, Network } from '@/types';
 
 export default Vue.extend({
   components: {
     Legend,
+  },
+
+  data(): {
+    searchTerm: string;
+    searchErrors: string[];
+    } {
+    return {
+      searchTerm: '',
+      searchErrors: [],
+    };
   },
 
   computed: {
@@ -119,6 +129,17 @@ export default Vue.extend({
         store.commit.setDirectionalEdges(value);
       },
     },
+
+    network(): Network | null {
+      return store.getters.network;
+    },
+
+    searchItems(): string[] {
+      if (this.network !== null) {
+        return this.network.nodes.map((node) => node._key);
+      }
+      return [];
+    },
   },
 
   methods: {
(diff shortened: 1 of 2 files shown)


jtomeck (dismissed):
This looks really good

src/components/Controls.vue
```diff
@@ -3,13 +3,23 @@ import Vue from 'vue';
...
 
+  data(): {
+    searchTerm: string;
+    searchErrors: string[];
+    } {
+    return {
+      searchTerm: '',
+      searchErrors: [],
+    };
+  },
```
waxlamp: Specifically for `data()` functions in Vue components, I like the following pattern a lot:

```suggestion
  data(): {
    return {
      searchTerm: '' as string,
      searchErrors: [] as string[],
    };
  },
```

This is one of the few times I use `as` without worrying about it.

src/components/Controls.vue
```diff
@@ -119,6 +129,17 @@ export default Vue.extend({
         store.commit.setDirectionalEdges(value);
       },
     },
+
+    network(): Network | null {
+      return store.getters.network;
+    },
+
+    searchItems(): string[] {
```
waxlamp: This function has an odd name. It's not searching the items, it's returning a candidate pool--is that right?

I'd suggest a better name for this function to avoid confusion for future devs.

src/components/Controls.vue
```diff
@@ -146,6 +167,20 @@ export default Vue.extend({
       a.click();
     },
 
+    search() {
+      const searchErrors: string[] = [];
+      if (this.network !== null) {
+        const nodeToSelect = this.network.nodes.find((node) => node._key === this.searchTerm);
```
waxlamp, line 170: We should keep an eye on this line of code in case it causes slowdowns with larger networks.

waxlamp (changes requested):
Nice work overall! I'd request you change the name of the `searchItems()` function; the other suggestions are at your discretion.

src/components/Controls.vue
JackWilb: Fixed in 30ff16a

waxlamp (approved):

PR #146, opened 2021-01-12
Refactor control panel width into store

### LINKED ISSUE #145: Refactor application so that the sidebar width is in a central location
The sidebar width needs to be tracked in the store, because it's referenced in several places. Some examples include:
- The sidebar component for the width
- The MultiLink component to determine offsets for the node drag and for the tooltip
- The Legend to set the max size

If we can store the width in one place, we can stop from hard coding it in the app.

JackWilb:
Closes #145

Refactors the control panel width into a variable that can be reused across components through the store.

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -123,10 +123,6 @@ export default Vue.extend({
     controlsWidth(): number {
       return store.getters.controlsWidth;
     },
-
-    navStyle(): string {
-      return `width: ${this.controlsWidth}px;`;
-    },
   },
 
   methods: {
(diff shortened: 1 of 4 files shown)


waxlamp (commented):
Code looks fine, but I'd consult with @jtomeck on whether there's a better way to handle this overall. It doesn't feel like the control panel width should be part of application state. Instead, it feels like there should be some other design protocol to handle this (very common) case of a sidebar and a main panel dividing the available horizontal space.

jtomeck (dismissed):
Looks fine to me. I'm not sure I have the best understanding of all the alternatives within a vuetify application, but this makes sense to me.

waxlamp (approved):

PR #151, opened 2021-01-14
Set the svg dimensions dynamically by using this.$vuetify to check for size changes

JackWilb:
This change allows the MultiLink svg to resize as the window changes size using a convenient vuetify binding. It also removes all the overflow and scroll bars that are there erroneously

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -80,6 +80,18 @@ export default {
 </template>
 
 <style>
+html {
+  scrollbar-width: none;
+}
+
+html::-webkit-scrollbar {
+  display: none;
+}
+
+body {
+  overflow: hidden;
+}
+
 #app {
   font-family: "Blinker", Helvetica, Arial, sans-serif;
   -webkit-font-smoothing: antialiased;
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -157,8 +157,8 @@ export default Vue.extend({
 
     svgDimensions() {
       return {
-        height: document.body.clientHeight,
-        width: document.body.clientWidth - 256,
+        height: this.$vuetify.breakpoint.height,
+        width: this.$vuetify.breakpoint.width - 256,
       };
     },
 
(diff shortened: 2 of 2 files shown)


src/App.vue
```diff
@@ -80,6 +80,18 @@ export default {
 </template>
 
 <style>
+html {
+  scrollbar-width: none;
+}
+
+html::-webkit-scrollbar {
```
waxlamp, line 87: I'm surprised you have to do any vendor-specific directives here. @jtomeck, do you know if there's a better way to handle this within a Vue style section?

src/components/MultiLink/MultiLink.vue
```diff
@@ -157,8 +157,8 @@ export default Vue.extend({
 
     svgDimensions() {
       return {
-        height: document.body.clientHeight,
-        width: document.body.clientWidth - 256,
+        height: this.$vuetify.breakpoint.height,
+        width: this.$vuetify.breakpoint.width - 256,
```
waxlamp: This looks pretty cool. I read the docs [here](https://vuetifyjs.com/en/features/breakpoints/#breakpoint-service) but I don't fully understand what `this.$vuetify` is meant for. I'm going to ask Jared for a formal review in part so we all can learn something new here.

waxlamp (dismissed):

src/components/MultiLink/MultiLink.vue
JackWilb: This use case, checking the size of the window (it will also give you which breakpoint we're in sm, md, lg, etc.) is for designing custom reactive elements that scale with the window. It's like how you can use a css framework for those breaks, but it gives you access to the info in javacript.

src/App.vue
JackWilb: I believe this was just best practice from what I could see. I'd be interested to see what @jtomeck has to say

src/components/MultiLink/MultiLink.vue
jtomeck: I think this is probably the best way to do this because of how vuetify handles responsive as well as css in general. What I mean is CSS breakpoints would probably not make sense.

jtomeck (dismissed):

PR #152, opened 2021-01-14
Update simulation center force as window resizes

JackWilb:
Depends on #151 

Adds a helper function to update the forces on the simulation, with the main use being that we can update the center force when the window resizes. I've set the helper to be automatically called as the `svgDimensions` variable updates.

### DIFF
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -180,8 +180,8 @@ export default Vue.extend({
       // Make the simulation
       const simulation = forceSimulation<Node, SimulationLink>()
         .force('center', forceCenter(this.svgDimensions.width / 2, this.svgDimensions.height / 2))
-        .force('charge', forceManyBody().strength(-300))
-        .force('link', forceLink().id((d) => { const datum = (d as Link); return datum._id; }))
+        .force('charge', forceManyBody<Node>().strength(-300))
+        .force('link', forceLink<Node, SimulationLink>().id((d) => { const datum = (d as Link); return datum._id; }))
         .force('collision', forceCollide(this.forceRadius));
 
       simulation
--- a/src/components/MultiLink/MultiLink.vue
+++ b/src/components/MultiLink/MultiLink.vue
@@ -150,6 +150,8 @@ export default Vue.extend({
       const { height } = this.$vuetify.breakpoint;
       const width = this.$vuetify.breakpoint.width - this.controlsWidth;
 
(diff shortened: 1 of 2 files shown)


src/components/MultiLink/MultiLink.vue
```diff
@@ -156,9 +156,14 @@ export default Vue.extend({
     },
 
     svgDimensions() {
+      const { height } = this.$vuetify.breakpoint;
+      const width = this.$vuetify.breakpoint.width - 256;
```
waxlamp: Where does the `256` come from?

src/store/index.ts
```diff
@@ -264,6 +266,16 @@ const {
     setDirectionalEdges(state, directionalEdges: boolean) {
       state.directionalEdges = directionalEdges;
     },
+
+    updateSimulationForce(state: State, payload: {
+      forceType: 'center' | 'charge' | 'link' | 'collision';
+      forceValue: ForceCenter<Node> | ForceManyBody<Node> | ForceLink<Node, SimulationLink> | ForceCollide<Node>;
+    }) {
```
waxlamp, line 281: Why a `payload` object, rather than `forceType` and `forceValue` (and why not simply `type` and `value`)?

waxlamp (commented):
Some questions to resolve before I approve.

src/components/MultiLink/MultiLink.vue
JackWilb: It's the width of the nav bar. I tried to address these magic numbers in #146

src/store/index.ts
JackWilb: It's a payload because the vuex mutations can only take one parameter (see [here](https://vuex.vuejs.org/guide/mutations.html#commit-with-payload)).

I called it `forceType` and `forceValue`, just to be explicit about what we're receiving. If you'd like me to rename them, that's fine. Should be a super quick fix.

waxlamp: > It's a payload because the vuex mutations can only take one parameter (see [here](https://vuex.vuejs.org/guide/mutations.html#commit-with-payload)).

Ah right, thanks for reminding me.

> I called it `forceType` and `forceValue`, just to be explicit about what we're receiving. If you'd like me to rename them, that's fine. Should be a super quick fix.

Totally up to you. Since the commit function has "force" in it, the word "force" seems redundant in the payload field names, but it really doesn't matter a lot.

JackWilb: Let's leave it as is for now

waxlamp (dismissed):

waxlamp (dismissed):

PR #155, opened 2021-01-15
Add remaining control panel elements to provenance (+ vis)

### LINKED ISSUE #97: Add more variables to provenance
Most things in the store should be tracked

JackWilb:
Depends on #156 for the fix to the v-selects

Closes #97

This tracks the remaining settings in the control panel in the provenance state. Most of them were pretty simple, aside from the 2 sliders. The sliders update their value constantly as they're dragged and that's undesirable for the provenance state (we instead want to know where the slider was dropped). Therefore I added a flag and a helper to update the provenance state for the sliders on `@change`.

There was a question of whether we should allow the MultiMatrix.vue component to modify the store directly (instead of through the vuex store), but I think that'll open us up to bugs and it means that the code for updating the provenance is dispersed throughout more files. This was the simplest change I could think of that kept all the relevant code together

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -76,7 +76,7 @@ export default Vue.extend({
         return store.getters.markerSize || 0;
       },
       set(value: number) {
-        store.commit.setMarkerSize(value);
+        store.commit.setMarkerSize({ markerSize: value, updateProv: false });
       },
     },
 
@@ -85,7 +85,7 @@ export default Vue.extend({
         return store.getters.fontSize || 0;
       },
       set(value: number) {
-        store.commit.setFontSize(value);
+        store.commit.setFontSize({ fontSize: value, updateProv: false });
       },
     },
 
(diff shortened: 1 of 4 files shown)


src/lib/provenanceUtils.ts
```diff
@@ -14,6 +14,36 @@ export function updateProvenanceState(vuexState: State, label: ProvenanceEventTy
...
+
+    if (label === 'Set Select Neighbors') {
+      // eslint-disable-next-line no-param-reassign
+      provState.selectNeighbors = newProvState.selectNeighbors;
+    }
+
+    if (label === 'Set Directional Edges') {
+      // eslint-disable-next-line no-param-reassign
+      provState.directionalEdges = newProvState.directionalEdges;
+    }
```
waxlamp: Using `else if` instead of parallel `if` statements would be good for two reasons:
1. prevents accidental insertion of code later between the statements
2. Makes clear to readers (via the if-else if-else idiom) that this is really a big "switch statement" (point 1 above is sort of a code enforcement of this idea).

src/components/Controls.vue
```diff
@@ -184,6 +180,16 @@ export default Vue.extend({
...
     },
+
+    updateSliderProv(value: number, type: 'markerSize' | 'fontSize') {
+      if (type === 'markerSize') {
+        store.commit.setMarkerSize({ markerSize: value, updateProv: true });
+      }
+
+      if (type === 'fontSize') {
+        store.commit.setFontSize({ fontSize: value, updateProv: true });
+      }
```
waxlamp: An `else if` construct would be good here too (see other comment).

waxlamp (dismissed):
Looks good. Take the if/else-if thing under advisement and if you decide to use that, you can self-approve and merge.

PR #156, opened 2021-01-15
Rework the v-select styling, the list logic for v-select items, and update default color value

JackWilb:
This is from a problem I noticed in #155 

The v-select elements were set to allow multiple selections and thus were returning arrays instead of just strings as expected. In order to fix the problem, I refactored the v-select style to not include `multiple`. I also added `clearable`, since we should be able to have no variable mapped to it, and removed unnecessary props: `chips`, `deletable-chips`, `small-chips`. and `persistent-hint`.

After dealing with that problem, I noticed some unnecessary logic in the logic that creates the list items and an incorrect default values for the color variable. As such, I removed the logic and update the default value to `''`

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -41,12 +41,8 @@ export default Vue.extend({
       return new Set();
     },
 
-    variableList(): Set<string | null> {
-      return this.multiVariableList.add(null);
-    },
-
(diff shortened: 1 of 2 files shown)


waxlamp (approved):
Nice work

PR #161, opened 2021-01-21
Add clear selection button with relevant provenance changes

JackWilb:
This was a feature requested by Alex. This button will eventually move to the always visible controls so I haven't put too much effort into styling it for now.

In the provenance, I added a check to see if we're trying to set an empty set for the selected nodes and if so, it adds a node to provenance called "Clear Selection"

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -160,6 +160,12 @@ const {
 
     setSelected(state, selectedNodes: Set<string>) {
       state.selectedNodes = selectedNodes;
+
+      if (state.provenance !== null) {
+        if (selectedNodes.size === 0) {
+          updateProvenanceState(state, 'Clear Selection');
+        }
+      }
     },
 
     setLoadError(state, loadError: LoadError) {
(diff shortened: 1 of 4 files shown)


src/store/index.ts
```diff
@@ -160,6 +160,12 @@ const {
 
     setSelected(state, selectedNodes: Set<string>) {
       state.selectedNodes = selectedNodes;
+
+      if (state.provenance !== null) {
+        if (selectedNodes.size === 0) {
+          updateProvenanceState(state, 'Clear Selection');
+        }
```
waxlamp, line 178: It seems like this condition will cause "clear selection" to appear in the provenance history even when simply de-selecting a single node. That's ok, but it's not precisely what the "clear selection" action means, so I'm mentioning here in case you didn't think of it.

waxlamp (approved):

src/store/index.ts
JackWilb: Adding and removing one node use a different function here, namely `addSelectedNode` and `removeSelectedNode` so the problem you mentioned shouldn't ever happen.

There could be another issue with the provenance here though, since this is used to reset the state when the provenance changes. I'll check and report back

JackWilb: And the provenance should be fine, it doesn't use these functions. It accesses the state directly with `storeState.selectedNodes = selectedNodes;` in store.ts

PR #169, opened 2021-01-22
Filter column attributes using type metadata

### LINKED ISSUE #158: Grab type data from the API and use it to choose which columns to visualize

JackWilb:
Depends on #159

Closes #158 

Removes and column with type label from the legend panel for data that have type.

I was trying to use `type: Set as PropType<Set<string>>` and it wasn't getting the type. My resolution was to just pass the type into the function, but that's not the most ideal way to get around the errors. Am I doing something wrong? Is there a better way to resolve this?

TODO:
- [x] Don't remove columns if the types don't exist (was already fixed because I'm checking !== 'label')
- [x] Show a message if no columns remain for visualization

### DIFF
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -61,7 +61,7 @@ export default Vue.extend({
     },
 
     columnTypes() {
-      const typeMapping: { [key: string]: string} = {};
+      const typeMapping: { [key: string]: string } = {};
 
       if (store.getters.networkMetadata !== null) {
         // eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -76,27 +76,11 @@ export default Vue.extend({
     },
 
     cleanedNodeVariables(): Set<string> {
-      const cleanedVariables = new Set<string>();
-
-      (this.multiVariableList as Set<string>).forEach((variable) => {
-        if (this.columnTypes[variable] !== 'label') {
-          cleanedVariables.add(variable);
-        }
-      });
-
-      return cleanedVariables;
+      return this.cleanVariableList(this.multiVariableList as Set<string>);
     },
 
     cleanedLinkVariables(): Set<string> {
-      const cleanedVariables = new Set<string>();
-
-      (this.linkVariableList as Set<string>).forEach((variable) => {
-        if (this.columnTypes[variable] !== 'label') {
-          cleanedVariables.add(variable);
-        }
-      });
-
-      return cleanedVariables;
+      return this.cleanVariableList(this.linkVariableList as Set<string>);
     },
   },
 
@@ -298,6 +282,18 @@ export default Vue.extend({
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     dragStart(newEvent: any) {
       newEvent.dataTransfer.setData('attr_id', newEvent.target.id);
+    },
+
+    cleanVariableList(list: Set<string>): Set<string> {
+      const cleanedVariables = new Set<string>();
+
+      list.forEach((variable) => {
+        if (this.columnTypes[variable] !== 'label') {
+          cleanedVariables.add(variable);
+        }
+      });
+
+      return cleanedVariables;
     },
   },
 });
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -502,6 +502,10 @@ export default Vue.extend({
         <br>
       </div>
 
+      <div v-if="cleanedNodeVariables.size === 0">
+        No Node Attributes To Visualize
+      </div>
+
       <br>
       <br>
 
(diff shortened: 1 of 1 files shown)


src/components/MultiLink/Legend.vue
```diff
@@ -58,6 +59,45 @@ export default Vue.extend({
     nodeColorScale() {
       return store.getters.nodeColorScale;
     },
+
+    columnTypes() {
+      const typeMapping: { [key: string]: string} = {};
```
waxlamp: There's a slight formatting issue on this line. Do you have prettier or eslint or something else running on this repo?

src/components/MultiLink/Legend.vue
```diff
@@ -58,6 +59,45 @@ export default Vue.extend({
...
+
+      if (store.getters.networkMetadata !== null) {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        Object.entries(store.getters.networkMetadata).forEach(([tableName, metadata]) => {
```
waxlamp: According to these [TS docs](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#array-destructuring), you can write this as:

```suggestion
        Object.entries(store.getters.networkMetadata).forEach(([, metadata]) => {
```

It's slightly ugly, but if this actually works (and the linter is ok with it) it's actually a better choice, because it prevents any errors from developing around the unused `tableName` variable.

waxlamp: Actually, if all you want is the object values, you should be able to get them directly with `Object.values()`, instead of calling `.entries()` and throwing away the first half of each element.

src/components/MultiLink/Legend.vue
```diff
@@ -58,6 +59,45 @@ export default Vue.extend({
...
+          (metadata as TableMetadata).table.columns.forEach((columnType) => {
+            typeMapping[columnType.key] = columnType.type;
+          });
```
waxlamp, line 70: Totally optional suggestion but this could be better expressed as a call to `.map()`.

src/components/MultiLink/Legend.vue
```diff
@@ -58,6 +59,45 @@ export default Vue.extend({
...
+      const cleanedVariables = new Set<string>();
+
+      (this.linkVariableList as Set<string>).forEach((variable) => {
+        if (this.columnTypes[variable] !== 'label') {
+          cleanedVariables.add(variable);
+        }
+      });
+
+      return cleanedVariables;
+    },
```
waxlamp: You may want to factor out the core of these two methods into a pure function.

src/components/MultiLink/Legend.vue
```diff
@@ -462,13 +502,17 @@ export default Vue.extend({
         <br>
       </div>
 
+      <div v-if="cleanedNodeVariables.size === 0">
+        No Node Attributes To Visualize
+      </div>
```
waxlamp: I think you should combine this with the v-for block above as its v-else. That is, display this block if there are no cleaned node variables; otherwise, display the v-for block. Might make this section easier to understand for the next person who comes along.

waxlamp: The titlecased message seems weird; can it just be capitalized normally?

waxlamp: (Same comments below, for the cleaned link variables.)

waxlamp (dismissed):
Looks good. Lots of suggestions; feel free to adopt the ones you want and ask me for a rubber stamp.

src/components/MultiLink/Legend.vue
JackWilb: Absolutely! `.values()` is a much better way to do this

src/components/MultiLink/Legend.vue
JackWilb: We're using eslint, but it doesn't complain about this line. Either way, I've fixed it in 0e719ac

src/components/MultiLink/Legend.vue
JackWilb: Fixed in c176361

src/components/MultiLink/Legend.vue
JackWilb: Fixed in 8f95444

src/components/MultiLink/Legend.vue
JackWilb: I had actually already done this before your review, I just forgot to push it. See a041160

src/components/MultiLink/Legend.vue
waxlamp: You should remove the comment directive.

src/components/MultiLink/Legend.vue
waxlamp: Turns out a map call here won't work; my bad.

src/components/MultiLink/Legend.vue
```diff
@@ -58,6 +59,29 @@ export default Vue.extend({
     nodeColorScale() {
       return store.getters.nodeColorScale;
     },
+
+    columnTypes() {
+      const typeMapping: { [key: string]: string } = {};
+
+      if (store.getters.networkMetadata !== null) {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
waxlamp:
```suggestion
```

waxlamp (changes requested):

src/components/MultiLink/Legend.vue
JackWilb: Thanks for the catch!

Issue #176, opened 2021-01-29
Variables in multiple tables with different types will cause a clash


JackWilb:
Currently, all code related to types implicitly expects each variable name to be unique for the type mapping. This assumption is probably not true, and in the case where we have the same variable name in different tables with different types, we'll have an issue.

waxlamp:
Fixing this could be as simple as keeping track of `<table name>/<column name>` instead of just `<column name>`, right?

JackWilb:
Yes. It would require changing to that around the app though so that we're always passing in that format

PR #177, opened 2021-01-29
Add UI and logic to set node color from the legend

### LINKED ISSUE #171: Make the legend reactive to displayCharts
When charts are displayed, show the mapping to bars and glyphs, else show node color and node size as the mapping options.

JackWilb:
Closes #171

This adds the remaining pieces of UI for adding a `nodeColorVairable` and a `nodeColorVariable` when the displayCharts variable is set to false. It removes the `v-select` from the controls panel in favor of this mechanism of setting the variables.

I also noticed that the legend was plotting far too many bars on each plot (one bar for each node/link), so I cleaned that up by using the `binLabels` as the bound data. That removes the duplicates.

We should open an issue to track potential performance issues from calculating size for each node on every tick of the simulation. We could refactor this code to write the size back to the node once and we'd read it from the node object. That would cut down on computation massively.

Through this work I noticed #176

### DIFF
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -167,7 +167,7 @@ export default Vue.extend({
           // Add the bars
           const variableSvgEnter = (variableSvg
             .selectAll()
-            .data(currentData)
+            .data(binLabels)
             .enter()
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             .append('rect') as any)
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -102,11 +102,11 @@ export default Vue.extend({
   methods: {
     setUpPanel() {
       // For node and link variables
-      [this.multiVariableList as Set<string>, this.linkVariableList as Set<string>].forEach((list) => {
+      [this.cleanedNodeVariables as Set<string>, this.cleanedLinkVariables as Set<string>].forEach((list) => {
         // For each attribute
         list.forEach((attr) => {
           // Get the SVG element and its width
-          const type = list === this.multiVariableList ? 'node' : 'link';
+          const type = list === this.cleanedNodeVariables ? 'node' : 'link';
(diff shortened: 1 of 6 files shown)


src/components/MultiLink/Legend.vue
```diff
@@ -226,7 +226,12 @@ export default Vue.extend({
       });
     },
 
+    // TODO: #176 use table name for var selection
```
waxlamp: The full link to GitHub would be better than just the `#176`.

src/components/MultiLink/Legend.vue
```diff
@@ -226,7 +226,12 @@ export default Vue.extend({
...
     isQuantitative(varName: string, type: 'node' | 'link') {
+      if (Object.entries(this.columnTypes).length > 0) {
```
waxlamp: Idiomatic usage would be to use `Object.keys()` here instead of `Object.entries()`.

This whole function is pretty odd--could you explain what it's doing? The heuristic "guess" at whether a variable is numeric after this block you've added makes me think there ought to be a Multinet-native way to do this.

waxlamp (dismissed):

src/components/MultiLink/Legend.vue
JackWilb: this function just if the multinet type of the variable is 'number' and returns true if it is. 

The heuristic is just for the data that is not typed in the database. Ultimately, we will probably want types on all data and the heuristic could be removed.

src/components/MultiLink/Legend.vue
JackWilb: See a081dfa

src/components/MultiLink/Legend.vue
JackWilb: See 7495a34 for `.entries()` -> `.keys()`

JackWilb: I'm going to go ahead and merge. Let's open an issue if you think this needs work

PR #178, opened 2021-01-29
Refactor the provenance visualization to use a right side nav-draw

### LINKED ISSUE #163: Move the trrack-vis to panel on the right (similar to the navigation bar)
Request from Alex

JackWilb:
Closes #163

Moves the provenance vis to a right side nav panel that can remain on the vis as the user is interacting with it. We could also reduce the size of the nodelink visualization area to just between the two panels when the provenance vis is rendering, but I'll check in with Alex to see his wishes for that.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,6 +7,7 @@ import {
 
 import Controls from './components/Controls.vue';
 import MultiLink from './components/MultiLink/MultiLink.vue';
+import ProvVis from './components/ProvVis.vue';
 
 export default {
   name: 'App',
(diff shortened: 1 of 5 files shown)


waxlamp (approved):
Code looks ok, though a lot of the state stuff seems like it could be avoided by using `v-model` on the `<v-navigation-drawer>` component. Just something to look at in case that's true.

JackWilb:
Since we need to toggle the state of the provenance menu from 2 locations, I think this is simplest. If it were just one, we could use the v-model

PR #179, opened 2021-02-02
Refactor the legend UI to show the values for selected variables

### LINKED ISSUE #142: Fix the sizing of legend drop targets
Remnant from #140

JackWilb:
Depends on #177

Closes #142

This refactors the UI for the legend drag targets and changes how the information is displayed. For visualized vars, show the categories/colors for each (including the mapping of colors to bars and glyphs).

Here's the current state:
Mapping node size and color:
image

Mapping to bars:
image

Mapping to bars and glyphs:
image

Mapping to link width (looks like the width scale needs some attention):
image

Link color also works, I just don't have a good example dataset for it.

The idea is to replicate some of the functionality from a past project. Here are some screen shots (note these are from a study to determine which visual elements were better so they vary and are inconsistent):

image

image

image

TODO:
- [x] Show color mapping for link variable
- [x] For link width show size mapping (#181)
- [x] Show color mapping for node color variable
- [x] For node size show size mapping (#180)
- [x] Refactor magic numbers
- [x] Move color swatches down and left (move onto separate row) (#183)
- [x] Add ability to remove a var from a mapping (#182)

### DIFF
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -33,14 +33,16 @@ export default Vue.extend({
     return {
       svgHeight: 50,
       yAxisPadding: 25, // Gives enough width for hundreds on the y axis
-      stickyBarHeight: 100,
-      stickyBarHorizSpacing: 60,
-      stickyBarWidth: 30,
-      stickyColorMapSquareSize: 15,
-      stickyPadding: 15,
-      stickyPlusBackgroundSize: 30,
-      stickyRowHeight: 50,
-      stickyVarNameIndent: 50,
+      sticky: {
+        barHeight: 100,
+        barHorizSpacing: 60,
+        barWidth: 30,
+        colorMapSquareSize: 15,
+        padding: 15,
+        plusBackgroundSize: 30,
+        rowHeight: 50,
+        varNameIndent: 50,
+      },
       varPadding: 10,
     };
   },
@@ -334,7 +336,7 @@ export default Vue.extend({
 
       const scale = scaleLinear()
         .domain([varMin, varMax])
-        .range([this.stickyBarHeight, 0]);
+        .range([this.sticky.barHeight, 0]);
       const axis = axisRight(scale).ticks(4, 's');
 
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


src/components/MultiLink/Legend.vue
```diff
@@ -33,6 +33,14 @@ export default Vue.extend({
...
       svgHeight: 50,
       yAxisPadding: 25, // Gives enough width for hundreds on the y axis
+      stickyBarHeight: 100,
+      stickyBarHorizSpacing: 60,
+      stickyBarWidth: 30,
+      stickyColorMapSquareSize: 15,
+      stickyPadding: 15,
+      stickyPlusBackgroundSize: 30,
+      stickyRowHeight: 50,
+      stickyVarNameIndent: 50,
```
waxlamp: Might it be easier to collect these into a single object named `sticky`?

src/components/MultiLink/Legend.vue
```diff
@@ -272,6 +300,9 @@ export default Vue.extend({
           glyph: this.nestedVariables.glyph,
         };
         store.commit.setNestedVariables(updatedNestedVars);
+
+        // Render the scales
+        Vue.nextTick(() => this.renderScales(droppedElText));
```
waxlamp, line 307: This feels a bit dangerous, though perhaps unavoidable. I think you could wrap the `renderScales()` method in a `watchEffect` if and when you refactor this to use composition API.

src/components/MultiLink/Legend.vue
```diff
@@ -297,6 +328,19 @@ export default Vue.extend({
...
+      const varMin = this.attributeRanges[barVar].min;
+      const varMax = this.attributeRanges[barVar].max;
+
+      const scale = scaleLinear()
+        .domain([varMin, varMax])
+        .range([this.stickyBarHeight, 0]);
+      const axis = axisRight(scale).ticks(4, 's');
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      select(`#node_${barVar}_scale`).call((axis as any));
```
waxlamp: Is this due to bad d3 typing?

waxlamp (dismissed):
Looks good overall. I would suggest refactoring many of the seemingly ad-hoc values in the template into computed props instead, especially if we want to reuse some of what you've created here in a toolkit of Multinet vis components. But no need to do that now, instead I'd file an issue as appropriate and defer to future work.

src/components/MultiLink/Legend.vue
JackWilb: See 3789364

src/components/MultiLink/Legend.vue
JackWilb: See #186

src/components/MultiLink/Legend.vue
JackWilb: It was. I managed to remove the `disable-next-line` by appending a g before calling the axis function on it in 6825a00

PR #185, opened 2021-02-04
Make the legend charts more compact and draw proper histograms for numeric variables

### LINKED ISSUE #172: Make legen charts more compact
Requested by Alex.

Reference lineup's implementation for an idea of size

JackWilb:
Closes #172 

This PR represents a bit of progression and regression. I made the legend more compact, used a better histogram function for the numeric variables, fixed the spacing around the axes, and made the charts responsive to the multinet types. However, I've also removed brushing (since it wasn't working and it'll require some bigger refactor to get it working).

I've opened #184 to track the regression and adding it back.

### DIFF
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -136,6 +136,36 @@ export default Vue.extend({
             }
 
             // TODO: add histogram code
+
+            xScale = scaleLinear()
+              .domain([min(currentData), max(currentData) + 1])
+              .range([this.yAxisPadding, variableSvgWidth]);
+
+            const binGenerator = histogram()
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              .domain((xScale as any).domain()) // then the domain of the graphic
+              .thresholds(xScale.ticks(15)); // then the numbers of bins
+
+            bins = binGenerator(currentData);
+
+            if (type === 'node') {
+              store.commit.addAttributeRange({ attr, min: parseFloat(min(currentData) || '0'), max: parseFloat(max(currentData) || '0') });
+            }
+
+            yScale = scaleLinear()
+              .domain([0, max(bins, (d) => d.length)])
+              .range([this.svgHeight, 0]);
+
+            variableSvg
+              .selectAll('rect')
+              .data(bins)
+              .enter()
+              .append('rect')
+              .attr('x', (d) => xScale(d.x0) || 0)
+              .attr('y', (d) => yScale(d.length) || 0)
+              .attr('height', (d) => this.svgHeight - yScale(d.length))
+              .attr('width', (d) => xScale(d.x1) - xScale(d.x0))
+              .attr('fill', (d) => (this.isQuantitative(attr, type) ? '#82B1FF' : this.nodeColorScale(d)));
           } else {
             if (type === 'node') {
               currentData = this.graphStructure.nodes.map((d: Node | Link) => d[attr]).sort();
--- a/src/components/MultiLink/Legend.vue
+++ b/src/components/MultiLink/Legend.vue
@@ -123,35 +123,57 @@ export default Vue.extend({
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           let currentData: any[];
 
-          if (type === 'node') {
-            currentData = this.graphStructure.nodes.map((d: Node | Link) => d[attr]);
+          let xScale: ScaleLinear<number, number> | ScaleBand<string> | null = null;
+          let yScale: ScaleLinear<number, number> | null = null;
+          let bins: Bin<number, number>[] | Map<string, number> | null = null;
+
+          // Process data for bars/histogram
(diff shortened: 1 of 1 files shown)


src/components/MultiLink/Legend.vue
```diff
@@ -28,8 +31,8 @@ export default Vue.extend({
 
   data() {
     return {
-      svgHeight: 150,
-      yAxisPadding: 10,
+      svgHeight: 50,
+      yAxisPadding: 25, // Gives enough width for hundreds on the y axis
```
waxlamp, line 35: Obviously, this will run into issues if you ever want to plot thousands, etc. We will eventually want a better way to decide how much padding to use; that's something @jtomeck might be able to help with.

src/components/MultiLink/Legend.vue
```diff
@@ -123,105 +126,98 @@ export default Vue.extend({
...
+
+          // Process data for bars/histogram
+          if (this.isQuantitative(attr, type)) {
+            if (type === 'node') {
+              currentData = this.graphStructure.nodes.map((d: Node | Link) => parseFloat(d[attr]));
+            } else {
+              currentData = this.graphStructure.edges.map((d: Node | Link) => parseFloat(d[attr]));
+            }
+
+            // TODO: add histogram code
```
waxlamp: Isn't that what this PR does?

src/components/MultiLink/Legend.vue
```diff
@@ -123,105 +126,98 @@ export default Vue.extend({
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
           let currentData: any[];
 
-          if (type === 'node') {
-            currentData = this.graphStructure.nodes.map((d: Node | Link) => d[attr]);
+          let xScale: ScaleLinear<number, number> | ScaleBand<string> | null = null;
```
waxlamp: In general, I think doing it like this defeats the point of using TypeScript. In the rest of the code, you are using `as` to force the type of `xScale` to what it needs to be in each branch. Ideally, you would just declare `xScale` with the correct type in each context in which it's used.

I'm not saying you have to stop and fix this now, but you should in more general terms slow down your use of `as` to dispatch between types the way you're doing here. Essentially, `as` should be used as an escape hatch when there are no other ways to proceed, and you have some out-of-band reasoning for why it won't cause a problem. (In this example, you *do* have such reasoning, but there may still be other ways to express the code.)

src/components/MultiLink/Legend.vue
```diff
@@ -123,105 +126,98 @@ export default Vue.extend({
...
+            bins = new Map([...new Set(currentData)].map(
+              (x) => [x, currentData.filter((y) => y === x).length],
+            )) as Map<string, number>;
+
+            const binLabels: string[] = [];
+            const binValues: number[] = [];
+            bins.forEach((value, label) => {
+              binLabels.push(label);
+              binValues.push(value);
+            });
```
waxlamp: Optional suggestion that increases legibility:

```suggestion
            const binLabels: string[] = bins.map((value, label) => label);
            const binValues: number[] = bins.map((value, label) => value);
```

waxlamp (dismissed):
Looks good!

In future, for PRs that change the look of the app, but that can't really be demoed via Netlify, could you include a screenshot or movie preview in the PR description?

src/components/MultiLink/Legend.vue
JackWilb: Yes, this was a remnant from cleaning up the commit stream. I left it in accidentally. It's removed in 8876d59

src/components/MultiLink/Legend.vue
JackWilb: I figured hundreds would be a safe value since for thousands, we'd need to load at least 1000 nodes, which is currently too much for the vis.

src/components/MultiLink/Legend.vue
JackWilb: You can't do `Map().map()` so, instead, I did:

```
            const binLabels: string[] = Array.from(bins.keys());
            const binValues: number[] = Array.from(bins.values());
```

See 2ff9b20

src/components/MultiLink/Legend.vue
waxlamp: Gotcha. Well, if it ever does scale that high, we can fix this then.

src/components/MultiLink/Legend.vue
waxlamp: Cool. Didn't realize `Map`s even had a `.forEach()` method.

waxlamp (dismissed):

src/components/MultiLink/Legend.vue
JackWilb: Agreed, I refactored this to use block scope vars inside the if and then I don't need to do all the forcing of types with `as`. I was trying to remove duplication of the axis appending code, but it makes sense for it to be duplicated -- they're affecting 2 different types of scales, numeric and categorical.

See 08c050c

Issue #166, opened 2021-01-21
Allow force directed layout of just some nodes, pinning others (by selection)


JackWilb:
Request from Alex

Relates to #164

PR #196, opened 2021-02-19
Add a context menu with a quick option to remove node selections

JackWilb:
This doesn't close an issue, but it's an improvement that we've wanted for a while. I added a context menu with a button to quickly clear the selection. I also used the composition API style and I feel like I'm becoming more familiar with it.

After reviewing the documentation, I saw there was an `@contextmenu` event hook so I'm using that. The only lingering issue was if the user held right click it would draw the rect select, so I made sure that only triggers on a left click by checking which button is pressed in the event handler.

This improvement will allow for a couple of improvements requested by Alex,  namely #166 and moving some of the menu buttons from the controls to this context menu (like I've done with "Clear Selection")

See #201 for a demo of the functionality

### DIFF
--- a/src/components/ContextMenu.vue
+++ b/src/components/ContextMenu.vue
@@ -1,0 +1,48 @@
+<script lang="ts">
+import store from '@/store';
+import { computed } from '@vue/composition-api';
+
+export default {
+  setup() {
+    const rightClickMenu = computed(() => store.getters.rightClickMenu);
+
+    function clearSelection() {
+      store.commit.setSelected(new Set());
+    }
+
+    return {
+      rightClickMenu,
+      clearSelection,
+    };
+  },
+};
+</script>
+
+<template>
+  <div
+    id="right-click-menu"
+  >
+    <v-menu
+      v-model="rightClickMenu.show"
+      :position-x="rightClickMenu.left"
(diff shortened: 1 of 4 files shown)


src/components/ContextMenu.vue
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+import store from '@/store';
+import { computed } from '@vue/composition-api';
+
+export default {
+  setup() {
+    const rightClickMenu = computed(() => store.getters.rightClickMenu);
```
waxlamp, line 7: It doesn't seem like this component should depend on values stored in global state. It seems a lot like this component ought to manage its own position values, and use a `v-model` pointing to a boolean stored in the component where the menu will appear. I found a nice article explaining how `v-model` works, and how it can be used in custom components: https://dev.to/vue-storefront/vue-during-coffee-break-using-v-model-with-custom-components-3bo9.

waxlamp (commented):
Looks nice and solid, though the use of global state to manage the visibility and position seems like an anti-pattern to me.

src/components/ContextMenu.vue
JackWilb: See https://github.com/multinet-app/multilink/issues/203

Issue #107, opened 2020-08-21
Give a more useful message when the network is too large to visualize


JackWilb:
Currently we just say "use AQL", which will be unsatisfactory for less technical users. See [comment](https://github.com/multinet-app/view-nodelink/pull/106#discussion_r474767456) for details.

(edited by @waxlamp to add): We should also consider whether redirecting the user to Multinet is the right thing to do. I think a better solution in general would be to allow the user to change some parameters in place: for a misspelled network name, we could just present a dropdown of existing networks. We could expand that to a general mechanism that allows for selecting both a workspace and a network. Then, the "sideloading" behavior would just be a special case of opening the app via query arguments. For "data too big" we could offer a ready made AQL query that samples the network down to the size limit by starting from a random or high-degree node. The associated error messages could then link to documentation for how to do this type of sampling, etc., yourself.

JackWilb:
We should use Derya's querying interface here when it's fully built

PR #205, opened 2021-02-27
Add dropdowns to alert to select workspace and graphs

JackWilb:
A partial solution to #107 (missing the AQL suggestion and the large network warnings)

Adds dropdowns to select a workspace and graph when the load fails because of a 404. We could possibly do something smarter with suggesting similar network names, but this does the simpler option -- letting the user choose.

The one question I have, is should we still offer a refresh on network error or should we show these dropdowns instead?

404 causes dropdowns:
image

404 dropdowns filled gives "GO TO GRAPH":
image

CORS error:
image

TODO:
- [x] Fix vertical padding
- [x] Add text hint under the main error message with resolution advice
- [x] Check what happens on permissions issue (give message to resolve it with the workspace owner)
- [x] Add `tile` prop to un-round the corners
- [x] Change graph to network (maybe this needs to be across this app)
- [x] If workspace/graph is missing, give a gentler message

### DIFF
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -63,6 +63,16 @@ export default {
       <v-row align="center">
         <v-col class="grow">
           {{ loadError.message }}
+
+          <br>
+
+          <small v-if="loadError.message === 'You are not authorized to view this workspace'">
+            If you are already logged in, please check with the workspace owner to verify your permissions.
+          </small>
+
+          <small v-else>
+            Select a workspace or network you'd like to return to.
+          </small>
         </v-col>
 
         <v-col
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -58,6 +58,7 @@ export default {
     <v-alert
       type="error"
       prominent
+      tile
     >
       <v-row align="center">
         <v-col class="grow">
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -59,7 +59,8 @@ export default {
 <template>
   <div>
     <v-alert
-      type="error"
+      type="warning"
+      border="left"
       prominent
       tile
     >
@@ -102,7 +103,12 @@ export default {
         </v-col>
 
         <v-col class="shrink">
-          <v-btn :href="buttonHref">
+          <v-btn
+            :href="buttonHref"
+            depressed
+            dark
+            color="grey darken-3"
+          >
             {{ buttonText }}
           </v-btn>
         </v-col>
--- a/src/components/Alert.vue
+++ b/src/components/Alert.vue
@@ -33,7 +33,7 @@ export default {
     watchEffect(async () => {
       if (workspace.value !== null && network.value !== null) {
         buttonHref.value = `./?workspace=${workspace.value}&graph=${network.value}`;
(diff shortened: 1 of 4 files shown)


waxlamp (changes requested):
- Change `graph` to `network`
- When the app is loaded without having specified a network, it should not behave as though something is wrong; instead, it should show a gentler message with the dropdowns directing the user to pick a workspace and network
- The CORS error scenario looks ok to me
- This is a case where it is more or less *required* to get @jtomeck's opinion on things--this is UX bread and butter and it's a mistake for us to proceed here without his expertise in the loop. I'd like for him to do a design review here, including getting an understanding of the problem we're trying to solve.

JackWilb:
Requested a review from @jtomeck. I'll wait for his feedback before I start addressing your points @waxlamp

jtomeck (changes requested):
1. I think on the alerts with dropdowns, the vertical padding is a little too much.
2. I also think adding instructions under the alert text in non-bold font like "select a workspace or network you'd like to return to." would be helpful
    - I think this also begs the question of why they're getting a "not found" error. Could that be a permissions issue? Maybe some text saying to contact the workspace owner or review permissions would be good here.
3. Can we also add the `tile` prop to the alert? It doesn't need rounded corners since it's flush to the edge.

JackWilb:
Here are all of the possible states:

Parameters missing from query:
image

Unauthorized:
image

Graph not found:
image

Workspace not found:
image

CORS error (any networking issue):
image

src/components/Alert.vue
```diff
@@ -0,0 +1,137 @@
...
+      workspaceOptions,
+    };
+  },
+};
+</script>
+
+<template>
+  <div>
+    <v-alert
+      type="error"
```
jtomeck:
```suggestion
      type="warning"
      outlined
      text
      border="left"
```

Roni mentioned in his comment that a softer message would be better and I think a softer color for the alert would probably be better as well.

src/components/Alert.vue
```diff
@@ -0,0 +1,137 @@
...
+                v-model="network"
+                label="Network"
+                :items="networkOptions"
+              />
+            </v-col>
+          </v-row>
+        </v-col>
+
+        <v-col class="shrink">
+          <v-btn :href="buttonHref">
```
jtomeck:
```suggestion
          <v-btn
              :href="buttonHref"
              depressed
          >
```

jtomeck (changes requested):
This is looking great, I added a couple code suggestions, but after that I think we're good.

jtomeck (dismissed):

JackWilb:
Here's the final design after a couple iterations with @jtomeck on slack.

image

waxlamp (changes requested):
This is looking great!

One big thing: this still uses the word "graph"--that needs to be "network".

JackWilb:
>One big thing: this still uses the word "graph"--that needs to be "network".

@waxlamp "Graph not found" comes from tha API. Instead of putting in custom logic here to change it, we should just fix it there

The button now says go to network:
image

waxlamp (approved):

PR #208, opened 2021-03-05
Add linkColorScale with support for numeric typed data

### LINKED ISSUE #57: Support continuous color maps for the link variables
See this [comment](https://github.com/multinet-app/view-nodelink/pull/51#discussion_r366495488) for the exact lines that will need to be updated.

JackWilb:
Closes #57

Adds a linkColorScale that uses type information to create a color scale for the links. If the variable is numeric, we use a sequential scale. If not, we use a ordinal scale. I refactored some type getting logic from the legend to the the store, since it will be reused across components.

The lingering todo is to show a representation of the color scale in the legend, but I'm defering that until after we're done with the re-design. I'll open an issue (see #209).

### DIFF
--- a/src/components/Legend.vue
+++ b/src/components/Legend.vue
@@ -8,7 +8,6 @@ import {
   scaleLinear, scaleBand, ScaleBand,
 } from 'd3-scale';
 import { axisBottom, axisLeft, axisRight } from 'd3-axis';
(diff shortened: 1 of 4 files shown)


JackWilb:
This is a relatively small change so I'm going to merge and we can iterate with more issues.

PR #213, opened 2021-03-17
Hide internal fields from label variable and from variables in the legend

### LINKED ISSUE #207: Hide internal fields from the label variables
_key shouldn't probably be allowed

JackWilb:
Closes #207

Hides internal fields from the UI of the application by using a typescript type and checking if the string is a member. This was the most elegant way I could think to do it without duplicating lots of the strings all over the application.

This is potentially breaking in the case where the `_key` variable is being used for storing useful info (like our les_mis d3 json example). The fix would be to warn users that this using the _key for anything other than IDs is unsupported, and change the d3 json parser to set an arbitrary key variable for json networks

This uses the `as string` type assertion, which I know is not the best. I couldn't figure out a way to tell typescript that this was a string. You can see that there are checks for undefined, but that doesn't seem to be letting typescript know that it must be string. Any help here would be great.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -108,3 +108,6 @@ export type ProvenanceEventTypes =
   'Set Node Size Variable' |
   'Set Select Neighbors'|
   'Set Directional Edges';
+
+export const internalFieldNames = ['_key', '_from', '_to', '_id', '_rev'] as const;
+export type InternalField = (typeof internalFieldNames)[number];
--- a/src/types.ts
+++ b/src/types.ts
@@ -72,7 +72,7 @@ export interface State {
   displayCharts: boolean;
   markerSize: number;
   fontSize: number;
-  labelVariable: string;
+  labelVariable: string | undefined;
   selectNeighbors: boolean;
   nestedVariables: NestedVariables;
   linkVariables: LinkStyleVariables;
--- a/src/types.ts
+++ b/src/types.ts
@@ -109,5 +109,5 @@ export type ProvenanceEventTypes =
   'Set Select Neighbors'|
   'Set Directional Edges';
 
-export const internalFieldNames = ['_key', '_from', '_to', '_id', '_rev'] as const;
+export const internalFieldNames = ['_from', '_to', '_id', '_rev'] as const;
 export type InternalField = (typeof internalFieldNames)[number];
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -3,7 +3,9 @@ import Vue from 'vue';
 import Legend from '@/components/Legend.vue';
 
 import store from '@/store';
-import { Node, Link, Network } from '@/types';
+import {
+  Node, Link, Network, internalFieldNames,
+} from '@/types';
 import { forceCollide, forceManyBody } from 'd3-force';
 
 export default Vue.extend({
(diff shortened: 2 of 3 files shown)


src/types.ts
```diff
@@ -108,3 +108,6 @@ export type ProvenanceEventTypes =
   'Set Node Size Variable' |
   'Set Select Neighbors'|
   'Set Directional Edges';
+
+export const internalFieldNames = ['_key', '_from', '_to', '_id', '_rev'] as const;
+export type InternalField = (typeof internalFieldNames)[number];
```
waxlamp, line 113: Took me a few minutes to learn this idiom (pretty cool). This [SO answer](https://stackoverflow.com/a/54061487/1886928) has it without the parentheses, which seems a bit clearer to me. But this is completely new to me, so I'm just curious where you obtained the syntax for this declaration. (Just trying to learn about this new idiom; I'm not making any specific suggestion for this line of code.)

waxlamp (commented):
As for your problems of typing `labelVariable`, I think you need to specify a return type of `string | undefined` [here](https://github.com/multinet-app/multilink/blob/d3422ff8ac0575df41651471e93c88cc6c4814c9/src/components/Controls.vue#L90). That's the value you're checking for `undefined`, and that would be the point where typescript would understand that its type has narrowed to `string`, and therefore you could dump the `as string` assertion.

More generally, I agree with you that the `_key` field can carry useful information, though not just in the case of Les Mis. I think being able to see the actual key value would be quite useful in case the dataset doesn't have a more natural text label to use, or if for any other reason a user would like to see it.

What do you think about removing `_key` from the list of banned columns?

src/types.ts
JackWilb: I saw this on SO, too. I wanted to have a list we could modify to select which variables are hidden and this was the main suggestion online

JackWilb:
> More generally, I agree with you that the `_key` field can carry useful information, though not just in the case of Les Mis. I think being able to see the actual key value would be quite useful in case the dataset doesn't have a more natural text label to use, or if for any other reason a user would like to see it.
> 
> What do you think about removing `_key` from the list of banned columns?

I think I've convinced myself that the `_key` variable should just be internal. Do you have some other examples of when we'd want to visualize the _key?

waxlamp:
> > More generally, I agree with you that the `_key` field can carry useful information, though not just in the case of Les Mis. I think being able to see the actual key value would be quite useful in case the dataset doesn't have a more natural text label to use, or if for any other reason a user would like to see it.
> > What do you think about removing `_key` from the list of banned columns?
> 
> I think I've convinced myself that the `_key` variable should just be internal. Do you have some other examples of when we'd want to visualize the _key?

We require a `_key` value as a generic identifier for each row of every table, and it would therefore act as an always-available fallback identifier in case there are no better label options. In fact, for a network that has nodes from multiple tables, even the `_id` value could be useful as such a fallback.

So, basically, these values are useful as labels when someone is just exploring their data and maybe looking at the table view in the main multinet client along with multilink to explore the topology of their network. What I'm suggesting is to put these back as valid options, but select them last in the guessing routine.

waxlamp (dismissed):
Apologies for getting to this so late.

I think this will be fine for now. We may need to remove `_from` and `_to` from the banned list at some point (under similar reasoning as for `_key`) but I don't think it's worth worrying about that now.

JackWilb:
Fixed up one small bug in 4a982f4. Going to merge with admin privileges since it doesn't change the functionality

PR #214, opened 2021-03-17
Automatically guess a label variable

JackWilb:
Depends on #213 

After removing `_key` from the options for label variable, it makes sense to try and guess a label variable for the dataset using the type information that we have available. To do so, I added another call to the end of the `fetchNetwork` function that sets a label variable based on the first variable of type label that it finds.

I'm not sure if this is the right place for the function, though, since it's not related to fetching the network. It's a bit of post processing once we have the data. Maybe it should live in App.vue after the fetchNetwork call happens. I'm open to suggestions

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -485,14 +485,6 @@ const {
 
       commit.setNetworkMetadata(networkMetadata);
       commit.setColumnTypes(networkMetadata);
-
-      // Guess the best label variable and set it
-      const allVars: Set<string> = new Set();
-      network.nodes.forEach((node: Node) => Object.keys(node).forEach((key) => allVars.add(key)));
-
-      const bestLabelVar = [...allVars]
-        .find((colName) => !isInternalField(colName) && context.getters.columnTypes[colName] === 'label');
-      commit.setLabelVariable(bestLabelVar);
     },
 
     releaseNodes(context) {
@@ -578,6 +570,18 @@ const {
           commit.startSimulation();
         }
       }
+    },
+
+    guessLabel(context) {
+      const { commit } = rootActionContext(context);
+
+      // Guess the best label variable and set it
+      const allVars: Set<string> = new Set();
+      context.getters.network.nodes.forEach((node: Node) => Object.keys(node).forEach((key) => allVars.add(key)));
+
+      const bestLabelVar = [...allVars]
+        .find((colName) => !isInternalField(colName) && context.getters.columnTypes[colName] === 'label');
+      commit.setLabelVariable(bestLabelVar);
     },
   },
 });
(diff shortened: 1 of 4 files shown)


src/store/index.ts
```diff
@@ -484,6 +485,14 @@ const {
 
       commit.setNetworkMetadata(networkMetadata);
       commit.setColumnTypes(networkMetadata);
+
+      // Guess the best label variable and set it
+      const allVars: Set<string> = new Set();
+      network.nodes.map((node: Node) => Object.keys(node).forEach((key) => allVars.add(key)));
```
waxlamp: Seems like this should be a `forEach()` rather than a `map()`.

src/store/index.ts
```diff
@@ -484,6 +485,14 @@ const {
...
+
+      const bestLabelVar = [...allVars]
+        .find((colName) => !isInternalField(colName) && context.getters.columnTypes[colName] === 'label');
+      commit.setLabelVariable(bestLabelVar);
```
waxlamp: I agree that this doesn't belong in here. It should go in a separate action, and that action should be invoked explicitly from its point of use. This is important in case, e.g., the user has already set a label field and then (for any reason) they need to re-fetch the network. You wouldn't want the user's choice to be overridden with this guess in that case.

waxlamp (changes requested):
If we restore `_key` to the list of allowed label variables, then the guessing routine would need to be modified a bit to make `_key` the lowest priority guess.

src/store/index.ts
JackWilb: See 4e5f193

src/store/index.ts
JackWilb: See 285650c

JackWilb:
> If we restore `_key` to the list of allowed label variables, then the guessing routine would need to be modified a bit to make `_key` the lowest priority guess.

Agreed, the fastest path forward is just not to use the _key. Otherwise, we'll need to add some logic that only selects it if it's the only label available. Some return logic like: 

```
if (labelOptions.length === 1) {
return '_key';
} else {
labelOptions.remove('_key');
```

waxlamp:
> > If we restore `_key` to the list of allowed label variables, then the guessing routine would need to be modified a bit to make `_key` the lowest priority guess.
> 
> Agreed, the fastest path forward is just not to use the _key. Otherwise, we'll need to add some logic that only selects it if it's the only label available. Some return logic like:
> 
> ```
> if (labelOptions.length === 1) {
> return '_key';
> } else {
> labelOptions.remove('_key');
> ```

This is not quite what I had in mind. I'm suggesting that the guessing routine should only select it if it's the only label available, but it should still be allowed as a label if the user wants it. (See the discussion on the other PR.)

JackWilb:
Rebased onto #213 again (after the _key change) and added logic to only select _key when it is the only candidate

waxlamp (dismissed):

JackWilb:
Updated to use `_key` last, as we had discussed

PR #216, opened 2021-03-29
Disable no-param-reassign for updateProvenanceState function

JackWilb:
This is a small update that reduces some of the needed syntax in the `updateProvenanceState` function block. I think this is more elegant and readable than doing it for every line


JackWilb:
Going to merge without review since it's a small change

PR #218, opened 2021-03-30
Track all variables that set forces in the store/provenance, update forces on provenance change

### LINKED ISSUE #193: Add simulation forces to provenance and track when they are updated
A couple of thoughts:

- Don't track on window resize
- Marker size could be an issue since it's already tracked
- Maybe linkLength should be refactored to be more like markerSize

JackWilb:
Closes #193

This PR adds support for tracking all variables that modify forces in the store and in the provenance. This gives users the ability to undo and redo the changes to the forces.

This PR does not track the forces that trigger on a window resize, since those shouldn't be undone and could cause problems with the simulation if they are reverted.

I noticed as I was making this PR that I'd need some common logic for applying the functions so I added that to a utils file and I just reference it where I need it. The function replaces another action function from the store, which would apply the forces. I'm not certain this was necessary, but it seemed a bit cleaner to me. You'll have to let me know your thoughts. 

*NOTE*: We had discussed an approach I was taking in the stand up on 3/30 where I outlined that all these parameters would be tracked in another object and then updated/restored by the provenance library. I have that code on a local branch and can share it, but it adds significant complexity and it would require some debugging with the trrack team. This branch seems like a much cleaner way to add this feature. Ping me if you'd like to see the alternative approach.

Video below:
https://user-images.githubusercontent.com/36867477/113050338-f3055400-9161-11eb-96ee-6ac3e7d4c962.mp4

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -8,7 +8,7 @@ import store from '@/store';
 import {
   Node, Link, Network, internalFieldNames,
 } from '@/types';
-import { forceCollide, forceManyBody } from 'd3-force';
+import { forceCollide } from 'd3-force';
 
 export default Vue.extend({
   components: {
@@ -21,7 +21,6 @@ export default Vue.extend({
     return {
       searchTerm: '' as string,
       searchErrors: [] as string[],
-      linkLength: '50',
       showTabs: false,
       tab: undefined,
     };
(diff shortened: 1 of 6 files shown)


JackWilb:
@waxlamp  I removed a request for you to view, since this is a minor change. I'll merge it as is

Issue #222, opened 2021-04-01
Add scales to the vis when laying out by numeric attributes


JackWilb:
Show the scale for x and y. Add some padding to the vis to put the scale in and to give it some breathing room.

JackWilb:
This is done

PR #223, opened 2021-04-01
Add logic to layout nodes by numeric attributes

### LINKED ISSUE #219: Add attribute driven layouts (numerical)
Nodes are placed in x and y by a numerical attribute.

JackWilb:
Closes #219

This adds the logic to layout the nodes by a node's numeric attributes. You can assign one variable to the x and one to the y. There is currently no feedback as to which is mapped where. Maybe this could be added to the legend in a later iteration.

I modified the logic here slightly to only accept variables that are node variables so there shouldn't be any issues.

@jtomeck, could you give me some pointers on UX here, please? It seems like there might be too many clicks

@waxlamp Is there a way for me to trigger a render update instead of requiring a node hover? Maybe using `Vue.forceUpdate`?

TODO:
- [x] Only accept node variables in the list of vars to layout by
- [x] Add issue to add axes to the vis (#222)
- [x] Consider whether we should track this in the legend, too. (#234)
- [x] Have to hover over node to cause update to trigger on second layout (#234)

Layout by 2 variables:
https://user-images.githubusercontent.com/36867477/114741076-aac06700-9d07-11eb-9830-ffe775f84ca4.mp4

Disabled numeric variables:
image

### DIFF
--- a/src/components/ContextMenu.vue
+++ b/src/components/ContextMenu.vue
@@ -7,25 +7,10 @@ export default {
     const rightClickMenu = computed(() => store.getters.rightClickMenu);
     const selectedNodes = computed(() => store.getters.selectedNodes);
     const network = computed(() => store.getters.network);
-    const numericVariables = computed(() => {
-      const numericColumns = Object.entries(store.getters.columnTypes)
-        .map(([key, value]) => (value === 'number' ? key : ''));
-
-      const numericColumnsSet = new Set(numericColumns);
-      numericColumnsSet.delete('');
-
-      numericColumnsSet.forEach((column) => {
-        if (
-          store.getters.network !== null
-          && store.getters.network.nodes[0] !== null
-          && store.getters.network.nodes[0][column] === undefined
-        ) {
-          numericColumnsSet.delete(column);
-        }
-      });
-
-      return [...numericColumnsSet].sort();
-    });
+    const numericVariables = computed(() => Object.entries(store.getters.columnTypes)
+      .filter(([, value]) => value === 'number')
+      .map(([key]) => key)
+      .sort());
 
     function clearSelection() {
       store.commit.setSelected(new Set());
@@ -140,7 +125,7 @@ export default {
                 v-bind="attrs"
                 v-on="on"
               >
-                Layout By
+                Lay Out By
                 <v-icon
                   dense
                   right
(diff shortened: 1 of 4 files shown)


waxlamp:
Is this ready for review or are you still working on it? If the latter, could you mark it as draft?

jtomeck:
@JackWilb I don't know how I missed this from 13 days ago! So sorry about that, I don't think it's too many clicks if you have the flyout menu for "Layout by" show on hover. Also I think "Layout by" should be changed to "Lay out by..."

If you want to meet to discuss further, feel free to throw something up on my calendar or hmu in slack

src/components/ContextMenu.vue
```diff
@@ -1,12 +1,31 @@
...
   setup() {
     const rightClickMenu = computed(() => store.getters.rightClickMenu);
     const selectedNodes = computed(() => store.getters.selectedNodes);
     const network = computed(() => store.getters.network);
+    const numericVariables = computed(() => {
+      const numericColumns = Object.entries(store.getters.columnTypes)
+        .map(([key, value]) => (value === 'number' ? key : ''));
+
+      const numericColumnsSet = new Set(numericColumns);
+      numericColumnsSet.delete('');
```
waxlamp: I think this is a classic use case for `.filter()` over `.map()`.

waxlamp (commented):
Roll with @jtomeck's review on this. I left one code improvement possibility that I can re-review if you choose to make it.

JackWilb:
Nested open-on-hover is a current feature request that open for vuetify. It's been open 4 years so I'm not sure when a resolution will be posted. In the thread for the issue there are some workarounds, but maybe we can just implement this in a different way. I think that'd be better than hacking this to work with the current implementation of vuetify. I'm open to suggestiongs

src/components/ContextMenu.vue
JackWilb: Fixed with 4d9fcef. Guess I got tunnel vision on this code block 😂

waxlamp:
@JackWilb, is there anything left to do here? Do you need another review from @jtomeck?

JackWilb:
From Jared on slack: 
>Yeah I don't think it's a dealbreaker, it just seems like a slightly better interaction. But I think you can soldier on and leave it as a click

I'll just wrap up the code problems and then we can look to merge

waxlamp (approved):

PR #225, opened 2021-04-06
Migrate settings and switches into the expandable tabs

### LINKED ISSUE #170: Simplify the menu
@alexsb Could you please fill this out with your desired design changes?

JackWilb:
Closes #170

This is the second PR for refactoring the sidebar after #224. These changes don't depend on it, though.his PR moves the settings and switches from the sidebar into a collasible menu with tabs. This should keep the controls within reach, but not take space wasy from the legend below.

I'd like some feedback here. I think we should put the remaining items into the menu, aside from node searching, but maybe that's too much. What do you all think about it?

You can see the changes in the deploy preview below.


jtomeck (changes requested):
This looks really good. I think the only thing I'd like to see is a little more padding on top and bottom of the menu. Especially when we move more of the tools back up into that area.

JackWilb:
> This looks really good. I think the only thing I'd like to see is a little more padding on top and bottom of the menu. Especially when we move more of the tools back up into that area.

Added a touch more padding to top and bottom of the cards in 06d8d6d. Let me know if that's what you meant @jtomeck

jtomeck:
> Let me know if that's what you meant @jtomeck

Perfect!

jtomeck (approved):

waxlamp (commented):
Code looks fine at a glance, so roll with Jared's reviews.

PR #284, opened 2021-10-20
Add axes to variable driven layouts

JackWilb:
Depends on #281
Closes (Looks like I didn't make this an issue...)

Add d3 axis to the bottom and left that show the values for nodes in a variable driven layout. I added some extra padding on the x and y scale in the index so that the nodes didn't overlap the axes.

There are also a couple small changes to how the variables are grouped together in the multilink file.

TODO:
- [x] Add axis labels
- [x] y-axis is flipped. Correct to lower values at the bottom
- [x] Layout is applied twice
- [x] Node values are off when the screen is resized
- [x] Fix misalignment of axes to node values (they're off an inconsistent amount for each tick, see image below)
image

### DIFF
--- a/src/components/ContextMenu.vue
+++ b/src/components/ContextMenu.vue
@@ -6,9 +6,8 @@ import {
 
 export default defineComponent({
   setup() {
-    const rightClickMenu = computed(() => store.state.rightClickMenu);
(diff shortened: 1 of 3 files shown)


JackWilb:
Here it is after the recent changes with a double encoded layout:

image

Issue #295, opened 2021-12-06
Break scales when the data has outliers


JackWilb:
Add a bin for "larger than <some value>"

JackWilb:
Potentially useful library https://github.com/lukewhyte/d3-scale-break

Seems like I'll just calculate IQR, and then make bins for high and low outliers based on the box and whisker plot algorithm. That should be robust enough for skewed data.

I'll also need to add the ability to switch the scale type (to log, for example), which will then need to use this code

PR #307, opened 2022-03-30
Move layout by variable UI to the legend from context menu

### LINKED ISSUE #286: Change interface for triggering an attribute driven layout

JackWilb:
### Does this PR close any open issues?
Closes #286

### Give a longer description of what this PR addresses and why it's needed
Moves the layout to the legend and uses the framework we have in place for dragging and dropping variables to assign them.

### Provide pictures/videos of the behavior before and after these changes (optional)
New look:
image

### Are there any additional TODOs before this PR is ready to go?

TODOs:
- [x] When there are 2 axes assigned and we remove only 1, it should switch back to the one axis layout
- [x] Check reactivity issues
- [x] Add metadata (like degree) (tracking elsewhere)
- [ ] #295 
- [x] #296

### DIFF
--- a/src/components/ContextMenu.vue
+++ b/src/components/ContextMenu.vue
@@ -8,27 +8,6 @@ export default defineComponent({
   setup() {
     const network = computed(() => store.state.network);
 
-    const numericVariables = computed(() => Object.entries(store.state.columnTypes || {})
-      .filter(([, value]) => value === 'number')
-      .map(([key]) => key)
-      .filter((key) => {
-        if (network.value !== null) {
-          return network.value.nodes[0][key] !== undefined;
-        }
-        return true;
-      })
-      .sort());
-    const categoricalVariables = computed(() => Object.entries(store.state.columnTypes || {})
-      .filter(([, value]) => value !== 'number' && value !== 'label')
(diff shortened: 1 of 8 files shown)


alexsb:
Feedback: Awesome! Really cool. 

Some thoughts: 

* Maybe we want a toggle or some way to explicitly switch to attribute driven layout? How is it implemented now, if an x and a y variable isn't defined, it uses forces? 

* Maybe we could use a better separation of layout/appearance variables and legend? Add a gray "Attributes" bar?

JackWilb:
> Feedback: Awesome! Really cool.
> 
> Some thoughts:
> 
>     * Maybe we want a toggle or some way to explicitly switch to attribute driven layout? How is it implemented now, if an x and a y variable isn't defined, it uses forces?
> 
>     * Maybe we could use a better separation of layout/appearance variables and legend? Add a gray "Attributes" bar?

I added a header for the attributes to give some visual separation.

As far as adding a switch, I think it might clutter up the UI. I think attribute driven layouts will be a common use case and having them always available isn't taking up too much real estate. I'm happy to reconsider if you feel strongly about it, though

PR #317, opened 2022-07-05
Use force directed layout for categorical variables

JackWilb:
### Does this PR close any open issues?
No

### Give a longer description of what this PR addresses and why it's needed
This PR makes the categorical layout use the force directed layout for jittering the nodes of one group. This means that the nodes are no longer overlapping each other when rendering in this manner.

### Provide pictures/videos of the behavior before and after these changes (optional)
I find this so satisfying:
https://user-images.githubusercontent.com/36867477/177649981-fe40bf4a-1be2-4852-959d-bd809b4facd3.mp4

### Are there any additional TODOs before this PR is ready to go?

TODOs:
- [x] Fix weird behavior with forces returning wrong value on first application
- [x] Fix charge force on closer layouts by removing it

### DIFF
--- a/src/components/MultiLink.vue
+++ b/src/components/MultiLink.vue
@@ -642,8 +642,6 @@ export default defineComponent({
         // Clear the label variable
         store.commit.setLabelVariable(undefined);
 
-        store.commit.stopSimulation();
-
         if (store.state.network !== null && store.state.columnTypes !== null) {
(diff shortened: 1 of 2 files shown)


alexsb:
This is awesome! One comment: the labels switch off automatically when switching to categorical layout. Is that intentional? It might not be necessary in this case.

